PR #11, opened 2023-10-03
Proposal: add dial constructors for all chainio structs

samlaf:
Been integrating the metrics and nodeapi with eigenDA, and it's a real pain to use our chain readers.
Sometimes I just need a simple chainreader, and I don't want to have to use the entire huge [constructor](https://github.com/Layr-Labs/eigensdk-go/blob/master/chainio/constructor/constructor.go).

Added one example of how we could do this, just to show what I'm thinking.

Wdyt @shrimalmadhur . Should we add these everywhere? I feel like it would be helpful.

### DIFF
--- a/chainio/elcontracts/reader.go
+++ b/chainio/elcontracts/reader.go
@@ -80,6 +80,30 @@ func NewELChainReader(
 	}, nil
 }
 
+// Dial follows the geth constructor name convention, and is a low level contructor that
+// builds a new ELChainReader from config values as opposed to other data types
+func Dial(
+	rpcHttpUrl string,
+	rpcWsUrl string,
+	slasherAddr gethcommon.Address,
+	blsPubKeyCompendiumAddr gethcommon.Address,
+	logger logging.Logger,
+) (*ELChainReader, error) {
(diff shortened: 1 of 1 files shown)


shrimalmadhur:
> Been integrating the metrics and nodeapi with eigenDA, and it's a real pain to use our chain readers. Sometimes I just need a simple chainreader, and I don't want to have to use the entire huge [constructor](https://github.com/Layr-Labs/eigensdk-go/blob/master/chainio/constructor/constructor.go).
> 
> Added one example of how we could do this, just to show what I'm thinking.
> 
> Wdyt @shrimalmadhur . Should we add these everywhere? I feel like it would be helpful.
> 
> ### Motivation
> ### Solution
> ### Open questions

I think we need to think about this from overall redesign perspective, based on usage which you are finding when using eigenda. because we will keep adding new kind of construction which will result in a very messy codebase. Let's discuss holistically how we want to design construction of clients.

samlaf:
I personally think it's better to have too many constructors than too few.

shrimalmadhur:
> I personally think it's better to have too many constructors than too few.

Ya I'm not against it. I think we just need to have a quick design session to figure out the correct patterns and then implement it.

PR #14, opened 2023-10-03
bug: fix bug in update operator details

shrimalmadhur:
Fixes # .

### Motivation
Fixing bug where operator update doesn't update delegation approver address

### Solution
Added mapping


samlaf (approved):
You were just forgetting to initialize that field? Should use a linter man, would have highlighted it in yellow 👯

PR #19, opened 2023-10-06
updated economic metrics to reflect new spec

samlaf:
Fixes # .

### DIFF
--- a/chainio/clients/avs_registry_contracts_client.go
+++ b/chainio/clients/avs_registry_contracts_client.go
@@ -144,10 +144,9 @@ func (a *AvsRegistryContractsChainClient) GetOperatorsStakeInQuorumsOfOperatorAt
 		blockNumber)
 	if err != nil {
 		return nil, nil, err
-	} else {
-		quorums := types.BitmapToQuorumIds(quorumBitmap)
-		return quorums, blsOperatorStateRetrieverOperator, nil
 	}
+	quorums := types.BitmapToQuorumIds(quorumBitmap)
+	return quorums, blsOperatorStateRetrieverOperator, nil
 }
 
 func (a *AvsRegistryContractsChainClient) GetCheckSignaturesIndices(
--- a/chainio/clients/avs_registry_contracts_client.go
+++ b/chainio/clients/avs_registry_contracts_client.go
@@ -8,6 +8,7 @@ import (
 	blspubkeyregistry "github.com/Layr-Labs/eigensdk-go/contracts/bindings/BLSPubkeyRegistry"
 	regcoord "github.com/Layr-Labs/eigensdk-go/contracts/bindings/BLSRegistryCoordinatorWithIndices"
 	"github.com/Layr-Labs/eigensdk-go/logging"
+	"github.com/Layr-Labs/eigensdk-go/types"
 
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	gethcommon "github.com/ethereum/go-ethereum/common"
@@ -29,9 +30,15 @@ type AVSRegistryContractsClient interface {
 
 	GetOperatorsStakeInQuorumsAtBlock(
 		opts *bind.CallOpts,
-		quorumNumbers []byte,
+		quorumNumbers []types.QuorumNum,
 		blockNumber uint32,
 	) ([][]blsoperatorstateretrievar.BLSOperatorStateRetrieverOperator, error)
+
+	GetOperatorsStakeInQuorumsOfOperatorAtBlock(
+		opts *bind.CallOpts,
+		operatorId types.OperatorId,
+		blockNumber uint32,
+	) ([]types.QuorumNum, [][]blsoperatorstateretrievar.BLSOperatorStateRetrieverOperator, error)
 
 	GetCheckSignaturesIndices(
 		opts *bind.CallOpts,
@@ -55,23 +62,23 @@ type AVSRegistryContractsClient interface {
 	) (*gethTypes.Transaction, error)
 }
 
-type AvsregistryContractsChainClient struct {
+type AvsRegistryContractsChainClient struct {
 	avsRegistryBindings     *avsRegistryContractBindings
 	ethHttpClient           eth.EthClient
 	registryCoordinatorAddr gethcommon.Address
 	logger                  logging.Logger
 }
 
-var _ AVSRegistryContractsClient = (*AvsregistryContractsChainClient)(nil)
+var _ AVSRegistryContractsClient = (*AvsRegistryContractsChainClient)(nil)
 
-func NewAVSContractsChainClient(
+func NewAvsRegistryContractsChainClient(
 	blsRegistryCoordinatorAddr gethcommon.Address,
 	blsOperatorStateRetrieverAddr gethcommon.Address,
 	stakeregistryAddr gethcommon.Address,
 	blsPubkeyRegistryAddr gethcommon.Address,
 	ethClient eth.EthClient,
 	logger logging.Logger,
-) (*AvsregistryContractsChainClient, error) {
+) (*AvsRegistryContractsChainClient, error) {
 	avsRegistryBindings, err := newAVSRegistryContractBindings(
 		blsRegistryCoordinatorAddr,
 		blsOperatorStateRetrieverAddr,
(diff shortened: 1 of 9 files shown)


metrics/collectors/economic.go
```diff
@@ -49,17 +48,33 @@ type EconomicCollector struct {
 	// then the operator's registeredStake will remain 600 until updateStakes() is called, at which point it will
 	// drop to the correct value of 300.
 	registeredStake *prometheus.Desc
+	
+	// TODO(samlaf): Removing this as not part of avs node spec anymore.
```
shrimalmadhur, line 52: do you still need this comment code?

metrics/collectors/economic.go
```diff
@@ -68,19 +83,19 @@ func NewEconomicCollector(
...
-			prometheus.Labels{},
-		),
 		registeredStake: prometheus.NewDesc(
 			"eigen_registered_stakes",
 			"Operator stake in <quorum> of <avs_name>'s StakeRegistry contract",
-			[]string{"quorum"},
+			[]string{"quorum_number", "quorum_name"},
 			prometheus.Labels{"avs_name": avsName},
 		),
+		// delegatedShares: prometheus.NewDesc(
```
shrimalmadhur, line 92: is it needed?

metrics/collectors/economic.go
```diff
@@ -91,8 +106,8 @@ func NewEconomicCollector(
 // descriptors.
 func (ec *EconomicCollector) Describe(ch chan<- *prometheus.Desc) {
 	ch <- ec.slashingStatus
-	ch <- ec.delegatedShares
 	ch <- ec.registeredStake
+	// ch <- ec.delegatedShares
```
shrimalmadhur, line 110: is it needed?

metrics/collectors/economic.go
```diff
@@ -116,28 +131,46 @@ func (ec *EconomicCollector) Collect(ch chan<- prometheus.Metric) {
...
+					stakeFloat64, _ := operator.Stake.Float64()
+					ch <- prometheus.MustNewConstMetric(
+						ec.registeredStake, prometheus.GaugeValue, stakeFloat64, strconv.Itoa(int(quorumNum)), ec.quorumNames[quorumNum],
+					)
+				}
+			}
+		}
+	}
+
+	// collect delegatedShares metric
```
shrimalmadhur, line 157: is this needed?

metrics/collectors/economic.go
```diff
@@ -116,28 +131,46 @@ func (ec *EconomicCollector) Collect(ch chan<- prometheus.Metric) {
...
-		}
-	}
-
 	// collect registeredStake metric
 	// TODO(samlaf): implement this. probably have to call the BLSOperatorStateRetriever contract?
 	// probably should start using the avsregistry service instead of chainio clients so that we can
 	// swap out backend for a subgraph eventually
+	quorumNums, blsOperatorStateRetrieverOperator, err := ec.avsRegistryReader.GetOperatorsStakeInQuorumsOfOperatorAtCurrentBlock(context.Background(), ec.operatorId)
+	if err != nil {
+		ec.logger.Error("Failed to get operator stake", "err", err)
```
shrimalmadhur, line 140: somehow I like early return instead of else
```
if err != nil {
  // do something
  return;
}
//do another thing
```
wdyt?

metrics/collectors/economic.go
```diff
@@ -116,28 +131,46 @@ func (ec *EconomicCollector) Collect(ch chan<- prometheus.Metric) {
...
+	} else {
+		for quorumIdx, quorumNum := range quorumNums {
+			// TODO: this is stupid.. when AVSs scale to have 5K operators we'll be running through a bunch of operators
```
shrimalmadhur, line 143: what is the reason of not doing this right now?

metrics/collectors/economic_test.go
```diff
@@ -10,28 +10,52 @@ import (
...
 	}
 
 	mockCtrl := gomock.NewController(t)
 	defer mockCtrl.Finish()
 
 	ethReader := mocks.NewMockELReader(mockCtrl)
 	ethReader.EXPECT().OperatorIsFrozen(gomock.Any(), operatorAddr).Return(false, nil)
-	ethReader.EXPECT().GetOperatorSharesInStrategy(gomock.Any(), operatorAddr, strategyAddrs[0]).Return(big.NewInt(1000), nil)
-	ethReader.EXPECT().GetOperatorSharesInStrategy(gomock.Any(), operatorAddr, strategyAddrs[1]).Return(big.NewInt(2000), nil)
+	// ethReader.EXPECT().GetOperatorSharesInStrategy(gomock.Any(), operatorAddr, strategyAddrs[0]).Return(big.NewInt(1000), nil)
```
shrimalmadhur, line 31: is it needed?

types/Operator.go
```diff
@@ -93,3 +93,18 @@ type OperatorAvsState struct {
 	StakePerQuorum map[QuorumNum]StakeAmount
 	BlockNumber    BlockNumber
 }
+
+var (
+	maxNumberOfQuorums = 192
+)
+
+func BitmapToQuorumIds(bitmap *big.Int) []QuorumNum {
```
shrimalmadhur, line 101: should it be in operator.go or some utils.go files?

chainio/clients/avs_registry_contracts_client.go
```diff
@@ -125,7 +132,25 @@ func (a *AvsregistryContractsChainClient) GetOperatorsStakeInQuorumsAtBlock(
...
+	blockNumber uint32,
+) ([]types.QuorumNum, [][]blsoperatorstateretrievar.BLSOperatorStateRetrieverOperator, error) {
+	quorumBitmap, blsOperatorStateRetrieverOperator, err := a.avsRegistryBindings.BlsOperatorStateRetriever.GetOperatorState0(
+		opts,
+		a.registryCoordinatorAddr,
+		operatorId,
+		blockNumber)
+	if err != nil {
+		return nil, nil, err
+	} else {
```
shrimalmadhur, line 147: you don't need an extra else here

metrics/collectors/economic.go
samlaf: Agree but can't do that here because I want to emit the other metrics (so can't return early).

metrics/collectors/economic.go
samlaf: Painful man.. I'm tired of implementing this crap, plus we only have 200 operators max so shouldn't be that big of a deal. If you force me to I can do it tomorrow :(

types/Operator.go
samlaf: I put it here because we otherwise have a circular dependency between utils and types...
this types package will require a refactor. I was thinking of putting your operator stuff in types/operator maybe, or something like that. Needs to be thought through how we want to separate utils and types. In some sense I'm thinking we should only have types here and all your operator methods (like Validate) should maybe be functions defined in utils?

chainio/clients/avs_registry_contracts_client.go
samlaf: thanks, fixed.

types/Operator.go
shrimalmadhur: the reason I added `Validate` is because it's a part of struct - as in you have to have `operator` object to call it. you can't call that without creating an object first.  It would be confusing to put it in utils. 
This function here doesn't require any object to be calling this - that's why it can added independently to utils. That's kind of my thought process in separating types vs utils. 

I feel we can have multiple utils based on package too.

metrics/collectors/economic.go
shrimalmadhur: are you talking about the commented code? I see.

samlaf: Ah I see your point. Indeed since its the last one I could do that. But I prefer to use the same structure across all the exported metrics. That way if I want to uncomment that code later, or add other metrics, I won't introduce a bug unknowingly.

types/Operator.go
samlaf: Don't understand your point. You can always turn a method into a function. Your Validate function could also just be written
```
Validate (operator, ...)
```

samlaf: In any case I can't move it to `utils.go` because that would create a circular dependency. What's your suggestion here?

shrimalmadhur: yea I didn't realize the circular dependency part. okay. I think it's fine here. we maybe have to think of clear separation.

`You can always turn a method into a function` - I think it's cleaner and very specific that validate is acted on the operator, where utils function can be more generic, like checking an eth address or a valid url vs utilities of our own types. We can think about that separation

shrimalmadhur (approved):

PR #20, opened 2023-10-06
separated rpc call metrics into a collector

samlaf:
this removes circular dependencies and makes it cleaner

Fixes # .

### DIFF
--- a/chainio/constructor/constructor.go
+++ b/chainio/constructor/constructor.go
@@ -28,13 +28,17 @@ type Config struct {
 	PromMetricsIpPortAddress      string `yaml:"prometheus_metrics_ip_port_address"`
 }
 
+// TODO: this is confusing right now because clients are not instrumented clients, but
+// we return metrics and prometheus reg, so user has to build instrumented clients at the call
+// site if they need them. We should probably separate into two separate constructors, one
+// for non-instrumented clients that doesn't return metrics/reg, and another instrumented-constructor
+// that returns instrumented clients and the metrics/reg.
 type Clients struct {
 	AvsRegistryChainReader avsregistry.AvsRegistryReader
 	AvsRegistryChainWriter avsregistry.AvsRegistryWriter
 	ElChainReader          elcontracts.ELReader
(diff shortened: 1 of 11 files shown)


metrics/eigenmetrics_example_test.go
```diff
@@ -74,8 +75,17 @@ func ExampleEigenMetrics() {
 	}
 	// We must register the economic metrics separately because they are exported metrics (from jsonrpc or subgraph calls)
 	// and not instrumented metrics: see https://prometheus.io/docs/instrumenting/writing_clientlibs/#overall-structure
-	economicMetricsCollector := collectors.NewEconomicCollector(eigenlayerReader, avsRegistryReader, "exampleAvs", logger, operatorAddr, quorumNames)
+	economicMetricsCollector := economic.NewCollector(eigenlayerReader, avsRegistryReader, "exampleAvs", logger, operatorAddr, quorumNames)
 	reg.MustRegister(economicMetricsCollector)
 
+	rpcCallsCollector := rpccalls.NewCollector("eigen", "exampleAvs", reg)
+	instrumentedEthClient, err := eth.NewInstrumentedClient("http://localhost:8545", rpcCallsCollector)
```
shrimalmadhur, line 82: this is nice!

shrimalmadhur (approved):

chainio/constructor/constructor.go
```diff
@@ -34,6 +34,7 @@ type Clients struct {
 	ElChainReader          elcontracts.ELReader
 	ElChainSubscriber      elcontracts.ELSubscriber
 	ElChainWriter          elcontracts.ELWriter
+	// TODO: should we return instrumented clients as well?
```
shrimalmadhur, line 37: umm maybe a separate constructor?

samlaf: Oh like a bare bone clients constructor (also wouldn't return metrics/reg), and then a instrumentedConstructor that returns instrumented clients and metrics/reg. I think that's a good idea, I'll update the comment. Don't want to implement right now though.

PR #21, opened 2023-10-06
fix key saving folder creation bug

shrimalmadhur:
Fixes # .

### Motivation
There was a bug in key saving logic where if the folder did not exist, we couldn't create keys. It was undetected because during manual testing the folder was there.

### Solution
Added a folder creation before saving to file and unit tested it.


crypto/bls/attestation.go
```diff
@@ -196,6 +198,11 @@ func (k *KeyPair) SaveToFile(path string, password string) error {
 		return err
 	}
 
+	dir := filepath.Dir(path)
+	if err := os.MkdirAll(dir, 0755); err != nil {
```
samlaf, line 202: Why is this 0755 but the writeFile below is 0644?

crypto/bls/attestation_test.go
```diff
@@ -0,0 +1,41 @@
...
+			password: "test",
+			wantErr:  false,
+		},
+	}
+
+	for name, tt := range tests {
+		t.Run(name, func(t *testing.T) {
+			t.Cleanup(func() {
+				dir := filepath.Dir(tt.keyPath)
+				_ = os.RemoveAll(dir)
```
samlaf, line 28: is ./operator_keys a path used for anything other than this test? If not this will delete potential user data.

crypto/bls/attestation_test.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+func TestBlsKeyCreation(t *testing.T) {
+	var tests = map[string]struct {
+		keyPath  string
+		password string
+		wantErr  bool
+	}{
+		"valid bls key save": {
+			keyPath:  "./operator_keys/test.bls.key.json",
```
samlaf, line 18: is ./operator_keys a path used for anything other than this test? If not this will delete potential user data.
Maybe rename to `operator_keys_test_dir` or something?

crypto/bls/attestation_test.go
shrimalmadhur: this is just for test - not used by any other tests.

crypto/bls/attestation_test.go
shrimalmadhur: oh ya can rename.

crypto/bls/attestation.go
shrimalmadhur: 755 is suited for folder creation 
644 is for non-executable file creation
https://www.maths.cam.ac.uk/computing/linux/unixinfo/perms#:~:text=755%20means%20you%20can%20do,users%20can%20only%20read%20it.

it doesn't matter in our case that much?

samlaf: I see. We might want to err on the more secure side and give 644 then? Why does that directory need to be executable..? Also probably don't want to give "others" the right to change my key file. Maybe 654 could work to give group access to write.. but if they really need that then they can go change it manually?

shrimalmadhur: Having execute permission on a directory authorizes you to look at extended information on files in the directory  like you can us `ls -l`

samlaf: We talked offline. Agree 755 for dirs and 644 for files is correct.

samlaf (approved):

samlaf (approved):

PR #22, opened 2023-10-06
fix eigen namespace

samlaf:
rpcCallsCollector was asking for eigen namespace, which doesn't make sense. This needs to be hardcoded in the sdk, so made a variable for that, and used it in metrics and 2 collectors.


types/misc.go
```diff
@@ -0,0 +1,3 @@
+package types
```
shrimalmadhur, line 1: can we name this file as constants.go and then can add more constants to single file

samlaf: done. I think at some point we should move it to a constants pkg maybe?
const.EigenPromNamespace is more explicit than types.EigenPromNamespace.
Will leave this for a future PR though. Maybe move `types/` and `consts/` under `common/` or something?

samlaf: Actually don't people use an `internal/` package for things that shouldn't be exposed? I feel like this eigen namespace shouldn't be exposed, so maybe put it there?

shrimalmadhur: Yea we could have something like internal if we don't have to expose it. But should be expose? If AVs wants to use that in any way?

shrimalmadhur (approved):

PR #23, opened 2023-10-10
add ecdsa utility methods

shrimalmadhur:
Fixes # .

Export methods to save and retrieve ecdsa keys

### DIFF
--- a/crypto/ecdsa/utils.go
+++ b/crypto/ecdsa/utils.go
@@ -12,11 +12,18 @@ import (
 	"github.com/google/uuid"
 )
 
+// WriteKey writes the private key to the given path
+// The key is encrypted using the given password
+// This function will create the directory if it doesn't exist
+// If there's an existing file at the given path, it will be overwritten
 func WriteKey(path string, privateKey *ecdsa.PrivateKey, password string) error {
 	UUID, err := uuid.NewRandom()
 	if err != nil {
 		return err
 	}
+
+	// We are using https://github.com/ethereum/go-ethereum/blob/master/accounts/keystore/key.go#L41
+	// to store the keys which requires us to have random UUID for encryption
 	key := &keystore.Key{
 		Id:         UUID,
 		Address:    crypto.PubkeyToAddress(privateKey.PublicKey),
--- a/crypto/ecdsa/utils.go
+++ b/crypto/ecdsa/utils.go
@@ -1,0 +1,73 @@
+package ecdsa
+
+import (
+	"bufio"
+	"crypto/ecdsa"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/ethereum/go-ethereum/accounts/keystore"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/google/uuid"
+)
+
+func WriteKey(path string, privateKey *ecdsa.PrivateKey, password string) error {
+	UUID, err := uuid.NewRandom()
+	if err != nil {
+		return err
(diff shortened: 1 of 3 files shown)


crypto/ecdsa/utils.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/google/uuid"
+)
+
+func WriteKey(path string, privateKey *ecdsa.PrivateKey, password string) error {
+	UUID, err := uuid.NewRandom()
+	if err != nil {
+		return err
+	}
+	key := &keystore.Key{
+		Id:         UUID,
```
samlaf, line 21: What is this UUID for? And why isn't it part of our privateKey struct? Seems weird that if we read/write/read/write a privateKey we end up writing with different UUIDs. Is it just useless? If so, add a comment

crypto/ecdsa/utils.go
```diff
@@ -0,0 +1,73 @@
...
+		return err
+	}
+
+	return writeBytesToFile(path, encryptedBytes)
+}
+
+func writeBytesToFile(path string, data []byte) error {
+	dir := filepath.Dir(path)
+	if err := os.MkdirAll(dir, 0755); err != nil {
+		fmt.Println("Error creating directories:", err)
```
samlaf, line 37: does `os.MkdirAll` error if the directory already exists? Weird to print an error like this if so.

crypto/ecdsa/utils.go
```diff
@@ -0,0 +1,73 @@
...
+		return err
+	}
+
+	file, err := os.Create(filepath.Clean(path))
+	if err != nil {
+		fmt.Println("file create error")
```
samlaf, line 43: this too

crypto/ecdsa/utils.go
shrimalmadhur: no if it's already created - it is a noop. - only if it's not able to create anything it gives error.

crypto/ecdsa/utils.go
shrimalmadhur: - UUID is required by the geth keystore encrypt function. it's just a unique identifier they use to store the keys 
 - we don't have a `privateKey` struct for ecdsa key. we have for BLS. here we are just using the crypto/ecdsa key type.
 - We are just providing a utility to store a private key this path. right the the caller of this is handling the override or not. we could introduce that validation in our library

crypto/ecdsa/utils.go
shrimalmadhur: this only created a file. This errors out if folder is not present which is why create before. This also errors out if it is not able to create file.

crypto/ecdsa/utils.go
samlaf: Sounds good. Please do add a comment with this information.

crypto/ecdsa/utils.go
samlaf: Ah Create's doc actually notes that the file will be overwritten (truncated) https://pkg.go.dev/os#Create
Can we also add a comment to this function saying to be careful that if there is already a key stored at that path it will be overwritten?

crypto/ecdsa/utils.go
shrimalmadhur: lol https://pkg.go.dev/os#MkdirAll says it's noop (If path is already a directory, MkdirAll does nothing and returns nil)  but not Mkdir

samlaf (approved):
LGTM

PR #24, opened 2023-10-10
Samlaf/bls aggregation service

samlaf:
Sorry for massive PR. This generic bls aggregation service turned out more complicated than I anticipated.
I think it still needs more tests, I'm prob forgetting some edge cases, so please go all out on the review.

### Motivation

Goal of this feature is to make it easy for avs teams to develop their own aggregator (basically bls signature aggregator).

### Solution

This service aggregates signatures over taskResponseDigests, and returns, after reaching the required signature threshold, all the information needed to verify the signature onchain (this [struct](https://github.com/Layr-Labs/eigenlayer-contracts/blob/dd27549ce4b6e0af6751f31a4b48ed643d3f0e19/src/contracts/middleware/BLSSignatureChecker.sol#L20)).

See the comment on top of the service definition for more info, but basically this service aggregates signatures by making the least number of assumption about task sending order. It's based off of the incredible-squaring avs task structure (which is similar to any offchain coprocessor/computation oracle system like truebit), which is that many tasks can be sent in parallel, and a task sent at a later time can be completed before a task sent earlier. This makes the goroutine play (one goroutine per task) necessary and complicates this service's implementation. A different service with a much more structured set of tasks (say eigenDA where tasks are only started offchain by aggregator->node messaging, and so aggregated task responses are always sent onchain orderly) could most likely benefit from a much simpler implementation, which we should eventually provide.

### Open questions

What do you guys think of the interface for this service? Open to suggestions to make this easier to use/more understandable. I added a bunch of comments and TODOs in the code listing the assumptions that this code is making, so please do go through them.


services/bls_aggregation/blsagg.go
```diff
@@ -0,0 +1,364 @@
...
+// map[taskResponseDigest]aggregatedOperators
+type aggregatedOperators struct {
+	// aggregate g2 pubkey of all operatos who signed on this taskResponseDigest
+	signersApkG2 *bls.G2Point
+	// aggregate signature of all operators who signed on this taskResponseDigest
+	signersAggSigG1 *bls.Signature
+	// aggregate stake of all operators who signed on this header for each quorum
+	signersTotalStakePerQuorum map[types.QuorumNum]*big.Int
+	// OperatorId of operators who signed on this header for each quorum
+	signersOperatorIdsSet map[types.OperatorId]bool
```
cryptogakusei, line 60: the type `map[types.OperatorId]bool` is only a map from an operatorID to a bool. There are no multiple quorums here.

samlaf: thanks, fixed.

services/bls_aggregation/blsagg.go
```diff
@@ -0,0 +1,364 @@
...
+		return <-signatureVerificationErrorC
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+}
+
+func (a *BlsAggregatorService) singleTaskAggregatorGoroutineFunc(
+	taskIndex types.TaskIndex,
+	taskCreatedBlock uint32,
+	quorumNumbers []types.QuorumNum,
```
cryptogakusei, line 179: in the comment to this function, it should be mentioned that the `quorumNumbers` contain identification of the quorums whose threshold percentages are defined in `quorumThresholdPercentages`. Without this comment, the customer of the SDK might miss on this relationship.

services/bls_aggregation/blsagg.go
```diff
@@ -0,0 +1,364 @@
...
+		totalStakePerQuorum[quorumNum] = quorumAvsState.TotalStake
+	}
+	quorumApksG1 := []*bls.G1Point{}
+	for _, quorumNumber := range quorumNumbers {
+		quorumApksG1 = append(quorumApksG1, quorumsAvsStakeDict[quorumNumber].AggPubkeyG1)
+	}
+
+	// TODO(samlaf): instead of taking a TTE, we should take a block as input
+	// and monitor the chain and only close the task goroutine when that block is reached
+	taskExpiredTimer := time.NewTimer(timeToExpiry)
```
cryptogakusei, line 210: reading your comment, how does the timer would read what is the current block number? Seems like there would be need for a timer routine which reads current block number which it then sends via chan to each of the active singleTaskAggregatorGoroutineFunc. wdyt?

services/bls_aggregation/blsagg.go
```diff
@@ -0,0 +1,364 @@
...
+	taskIndex types.TaskIndex,
+	taskCreatedBlock uint32,
+	quorumNumbers []types.QuorumNum,
+	quorumThresholdPercentages []types.QuorumThresholdPercentage,
+	timeToExpiry time.Duration,
+) error {
+	if _, taskExists := a.taskChans[taskIndex]; taskExists {
+		return TaskAlreadyInitializedErrorF(taskIndex)
+	}
+	taskC := make(chan types.SignedTaskResponseDigest)
```
cryptogakusei, line 137: why not name `taskC` as `taskRespC`? It is named as such in `singleTaskAggregatorGoroutineFunc`.

services/bls_aggregation/blsagg_test.go
```diff
@@ -0,0 +1,327 @@
...
+			NonSignersPubkeysG1: []*bls.G1Point{},
+			QuorumApksG1:        []*bls.G1Point{testOperator1.BlsKeypair.GetPubKeyG1()},
+			SignersApkG2:        testOperator1.BlsKeypair.GetPubKeyG2(),
+			SignersAggSigG1:     testOperator1.BlsKeypair.SignMessage(taskResponseDigest1),
+		}
+		gotAggregationServiceResponse := <-blsAggServ.ResponseC
+		require.Equal(t, wantAggregationServiceResponse, gotAggregationServiceResponse)
+	})
+
+	t.Run("1 quorum 2 operator 2 signatures on 2 different msgs - task expired", func(t *testing.T) {
```
cryptogakusei, line 283: it seems we don't have a test for two different concurrently active tasks

services/bls_aggregation/blsagg.go
samlaf: Added bunch of comments about this [0b6351d](https://github.com/Layr-Labs/eigensdk-go/pull/24/commits/0b6351d05b09cb7e98a3750d8822b36d27b7bc0b)

services/bls_aggregation/blsagg.go
samlaf: You'd just set a ticker that ticks every 12s (roughly once per block), queries the chain for blockNum, check if its reached the TTEBlock, and if so close the goroutine.

services/bls_aggregation/blsagg.go
samlaf: Good catch. Updated channel names to aggregatedResponsesC and signedTaskRespsCs: [7172ee2](https://github.com/Layr-Labs/eigensdk-go/pull/24/commits/7172ee23384a8921450b45c8c8c3d123b2ad6422)

services/bls_aggregation/blsagg_test.go
samlaf: Indeed. Need to add this and test with >1 quorums

crypto/bls/aggregation.go
```diff
@@ -134,7 +134,7 @@ func (a *StdSignatureAggregator) AggregateSignatures(
 				aggSigs[ind] = &Signature{sig.Deserialize(sig.Serialize())}
 				aggPubKeys[ind] = op.PubkeyG2.Deserialize(op.PubkeyG2.Serialize())
 			} else {
-				aggSigs[ind].Add(sig.G1Point)
+				aggSigs[ind].Add(sig)
```
shrimalmadhur, line 137: I am curious why has this changed?

crypto/bls/attestation.go
```diff
@@ -47,14 +47,20 @@ func NewG1Point(x, y *big.Int) *G1Point {
 	}
 }
 
+func NewZeroG1Point() *G1Point {
+	return NewG1Point(big.NewInt(0), big.NewInt(0))
+}
+
 // Add another G1 point to this one
-func (p *G1Point) Add(p2 *G1Point) {
+func (p *G1Point) Add(p2 *G1Point) *G1Point {
```
shrimalmadhur, line 55: What is the point of returning it if we are not creating a new object?

crypto/bls/attestation.go
```diff
@@ -112,6 +124,15 @@ type Signature struct {
 	*G1Point `json:"g1_point"`
 }
 
+func NewZeroSignature() *Signature {
+	return &Signature{NewZeroG1Point()}
+}
+
+func (s *Signature) Add(otherS *Signature) *Signature {
+	s.G1Point.Add(otherS.G1Point)
```
shrimalmadhur, line 132: same thing.

services/avsregistry/avsregistry_chaincaller.go
```diff
@@ -12,9 +12,11 @@ import (
 	"github.com/Layr-Labs/eigensdk-go/types"
 )
 
+// AvsRegistryService is a wrapper around AvsRegistryReader that transforms the data into
```
shrimalmadhur, line 15: `AvsRegistryService` --> `AvsRegistryServiceChainCaller`

types/operator.go
```diff
@@ -83,15 +83,16 @@ type StakeAmount = *big.Int
 // It is currently the hash of the operator's G1 pubkey (in the bls pubkey registry)
 type OperatorId = [32]byte
 type QuorumNum = uint8
-type BlockNumber = uint32
+type QuorumThresholdPercentage = uint32
+type BlockNum = uint32
```
shrimalmadhur, line 87: Why can't we keep the type `BlockNumber` ?

services/avsregistry/avsregistry_chaincaller_test.go
```diff
@@ -144,15 +140,85 @@ func TestAvsRegistryServiceChainCaller_GetOperatorsAvsState(t *testing.T) {
...
+	testOperator := testOperator{
+		operatorAddr: common.HexToAddress("0x1"),
+		operatorId:   types.OperatorId{1},
+		pubkeys: types.OperatorPubkeys{
+			G1Pubkey: bls.NewG1Point(big.NewInt(1), big.NewInt(1)),
+			G2Pubkey: bls.NewG2Point([2]*big.Int{big.NewInt(1), big.NewInt(1)}, [2]*big.Int{big.NewInt(1), big.NewInt(1)}),
+		},
+	}
+
+	var tests = []struct {
```
shrimalmadhur, line 165: Do you need to add any error test cases?

services/avsregistry/avsregistry_fake.go
```diff
@@ -0,0 +1,72 @@
...
+	blsoperatorstateretrievar "github.com/Layr-Labs/eigensdk-go/contracts/bindings/BLSOperatorStateRetriever"
+	"github.com/Layr-Labs/eigensdk-go/crypto/bls"
+	"github.com/Layr-Labs/eigensdk-go/types"
+)
+
+type FakeAvsRegistryService struct {
+	operators map[types.BlockNum]map[types.OperatorId]types.OperatorAvsState
+}
+
+func NewFakeAvsRegistryService(blockNum types.BlockNum, operators []types.TestOperator) *FakeAvsRegistryService {
```
shrimalmadhur, line 17: Where is this used?

services/gen.go
```diff
@@ -2,3 +2,12 @@ package services
 
 //go:generate mockgen -destination=./mocks/pubkeycompendium.go -package=mocks github.com/Layr-Labs/eigensdk-go/services/pubkeycompendium PubkeyCompendiumService
 //go:generate mockgen -destination=./mocks/avsregistry.go -package=mocks github.com/Layr-Labs/eigensdk-go/services/avsregistry AvsRegistryService
+
+// We generate it in ./mocks/blsagg/ instead of ./mocks like the others because otherwise we get a circular dependency
+// avsregistry -> mocks -> avsregistry
+// because avsregistry_chaincaller_test -> for pubkeycompendium mock
+// and blsaggregation mock -> avsregistry interface
+// TODO: are there better ways to organize these dependencies? Maybe by using ben johnson
```
shrimalmadhur, line 10: yea might be a better idea to put mocks in the packages itself then in the uber mocks in`services`

types/avs.go
```diff
@@ -0,0 +1,20 @@
+package types
+
+import "github.com/Layr-Labs/eigensdk-go/crypto/bls"
+
+type TaskIndex = uint32
+type TaskDict struct {
+	// We only use a single quorum for incredible squaring
```
shrimalmadhur, line 7: So we can't use this for multiquroums? This will change once we have that functionality right? Can we add explicitly a TODO to change this later?

types/operator.go
```diff
@@ -108,3 +109,10 @@ func BitmapToQuorumIds(bitmap *big.Int) []QuorumNum {
 	}
 	return quorumIds
 }
+
+type QuorumAvsState struct {
```
shrimalmadhur, line 113: I think instead of loading one file with all the struct we should create new files to keep it cleaner. This is in the `operator` but seems it has relations AVS state?  This file is already getting too big for operator types.

crypto/bls/aggregation.go
samlaf: Because Signatures wrap a G1Point, and the library originally didn't overwrite the Add method, so that add method being used was signatures was actually the Add method of the underlying G1Point. I found it ugly that you needed to do `sig.Add(othersig.G1point)`, so added the method on signatures directly so that you just have to do `sig.Add(otherSig)`

crypto/bls/attestation.go
samlaf: This way you can chain Adds. Instead of having to do:
```
p1.Add(p2)
p1.Add(p3)
```
you can just do
```
p1.Add(p2).Add(p3)...
```

see for eg: https://github.com/Layr-Labs/eigensdk-go/blob/7172ee23384a8921450b45c8c8c3d123b2ad6422/services/bls_aggregation/blsagg_test.go#L103

crypto/bls/attestation.go
samlaf: see other response.

services/bls_aggregation/blsagg_test.go
samlaf: Added them in [cdd35f8](https://github.com/Layr-Labs/eigensdk-go/pull/24/commits/cdd35f8c99748ef2225fbae231581522297e46cb) and [c02f604](https://github.com/Layr-Labs/eigensdk-go/pull/24/commits/c02f6044d56c8b9e4e2d183c3aef9b1a2359e989)

services/avsregistry/avsregistry_chaincaller.go
samlaf: Fixed: [0e3d89a](https://github.com/Layr-Labs/eigensdk-go/pull/24/commits/0e3d89a9e7586426d0b78238acd8efdb6865a353)

types/operator.go
samlaf: Changed it to `BlockNum` so it fits with `QuorumNum`

services/avsregistry/avsregistry_chaincaller_test.go
samlaf: Probably should eventually, but no time right now. Added TODO in: [9066456](https://github.com/Layr-Labs/eigensdk-go/pull/24/commits/90664564eb4cb69cc1b500917defdf090c84e6d4)

services/avsregistry/avsregistry_fake.go
samlaf: used in all the `blsagg_test.go` test cases. I'm using this instead of a mock, as we discussed.

services/gen.go
samlaf: We can clean this later.

types/avs.go
samlaf: Good catch. This struct is no longer used. Deleted: [b889aa3](https://github.com/Layr-Labs/eigensdk-go/pull/24/commits/b889aa3b3930741bc4771b9f078871743439e595)

types/operator.go
samlaf: I know... we need a proper cleanup of types as we've discussed a few times. Can we do this in another ticket? I need this merged to be able to release inc-sq

services/bls_aggregation/blsagg_test.go
cryptogakusei: I just skimmed over them. Seems good. Resolving conversation.

types/operator.go
shrimalmadhur: i have created an issue #42

services/bls_aggregation/blsagg.go
```diff
@@ -0,0 +1,387 @@
...
+	"time"
+
+	"github.com/Layr-Labs/eigensdk-go/crypto/bls"
+	"github.com/Layr-Labs/eigensdk-go/logging"
+	"github.com/Layr-Labs/eigensdk-go/services/avsregistry"
+	"github.com/Layr-Labs/eigensdk-go/types"
+)
+
+var (
+	TaskAlreadyInitializedErrorF = func(taskIndex types.TaskIndex) error {
```
shrimalmadhur, line 18: `TaskAlreadyInitializedErrorFn`

services/bls_aggregation/blsagg.go
```diff
@@ -0,0 +1,387 @@
...
+		return fmt.Errorf("task %d already initialized", taskIndex)
+	}
+	TaskExpiredError   = fmt.Errorf("task expired")
+	TaskNotFoundErrorF = func(taskIndex types.TaskIndex) error {
+		return fmt.Errorf("task %d not initialized or already completed", taskIndex)
+	}
+	OperatorNotPartOfTaskQuorumErrorF = func(operatorId types.OperatorId, taskIndex types.TaskIndex) error {
```
shrimalmadhur, line 25: OperatorNotPartOfTaskQuorumErrorFn

services/bls_aggregation/blsagg.go
```diff
@@ -0,0 +1,387 @@
...
+		return fmt.Errorf("operator %x not part of task %d's quorum", operatorId, taskIndex)
+	}
+	SignatureVerificationError = func(err error) error {
```
shrimalmadhur, line 28: `SignatureVerificationErrorFn`

services/bls_aggregation/blsagg.go
```diff
@@ -0,0 +1,387 @@
...
+	_, ok := operatorsAvsStateDict[signedTaskResponseDigest.OperatorId]
+	if !ok {
+		a.logger.Warnf("Operator %#v not found. Skipping message.", signedTaskResponseDigest.OperatorId)
+		return OperatorNotPartOfTaskQuorumErrorF(signedTaskResponseDigest.OperatorId, taskIndex)
+	}
+
+	// 0. verify that the msg actually came from the correct operator
+	operatorG2Pubkey := operatorsAvsStateDict[signedTaskResponseDigest.OperatorId].Pubkeys.G2Pubkey
+	if operatorG2Pubkey == nil {
+		a.logger.Fatalf("Operator G2 pubkey not found")
```
shrimalmadhur, line 340: can just use `Fatal` if not use string formatting

services/bls_aggregation/blsagg.go
```diff
@@ -0,0 +1,387 @@
...
+		"taskResponseDigest", signedTaskResponseDigest.TaskResponseDigest,
+		"blsSignature", signedTaskResponseDigest.BlsSignature,
+	)
+	signatureVerified, err := signedTaskResponseDigest.BlsSignature.Verify(operatorG2Pubkey, signedTaskResponseDigest.TaskResponseDigest)
+	if err != nil {
+		a.logger.Errorf(SignatureVerificationError(err).Error())
+		return SignatureVerificationError(err)
+	}
+	if !signatureVerified {
+		a.logger.Errorf(IncorrectSignatureError.Error())
```
shrimalmadhur, line 353: can just use `Error` if not use string formatting

services/bls_aggregation/blsagg.go
samlaf: [3a8dd29](https://github.com/Layr-Labs/eigensdk-go/pull/24/commits/3a8dd2957ca7dfe8694773728008e7cca5eb60c1)

services/bls_aggregation/blsagg.go
samlaf: [3a8dd29](https://github.com/Layr-Labs/eigensdk-go/pull/24/commits/3a8dd2957ca7dfe8694773728008e7cca5eb60c1)

services/bls_aggregation/blsagg.go
samlaf: [3a8dd29](https://github.com/Layr-Labs/eigensdk-go/pull/24/commits/3a8dd2957ca7dfe8694773728008e7cca5eb60c1)

services/bls_aggregation/blsagg.go
```diff
@@ -0,0 +1,387 @@
...
+		a.logger.Fatalf("Operator G2 pubkey not found")
+	}
+	a.logger.Debug("Verifying signed task response digest signature",
+		"operatorG2Pubkey", operatorG2Pubkey,
+		"taskResponseDigest", signedTaskResponseDigest.TaskResponseDigest,
+		"blsSignature", signedTaskResponseDigest.BlsSignature,
+	)
+	signatureVerified, err := signedTaskResponseDigest.BlsSignature.Verify(operatorG2Pubkey, signedTaskResponseDigest.TaskResponseDigest)
+	if err != nil {
+		a.logger.Errorf(SignatureVerificationError(err).Error())
```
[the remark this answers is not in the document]
samlaf: [3a8dd29](https://github.com/Layr-Labs/eigensdk-go/pull/24/commits/3a8dd2957ca7dfe8694773728008e7cca5eb60c1)

services/bls_aggregation/blsagg.go
samlaf: [3a8dd29](https://github.com/Layr-Labs/eigensdk-go/pull/24/commits/3a8dd2957ca7dfe8694773728008e7cca5eb60c1)

samlaf:
@shrimalmadhur addressed all your comments. Should be good to go!

shrimalmadhur (approved):

PR #32, opened 2023-10-11
fix for delegation approver address

### LINKED ISSUE #29: Fix validation of operator configuration - DelegationApproverAddress
**Describe the bug**

Right now the validation of operator is not correct. According to EL [contracts](https://github.com/Layr-Labs/eigenlayer-contracts/blob/master/src/contracts/interfaces/IDelegationManager.sol#L28) It can be set to `0x0` 

https://github.com/Layr-Labs/eigensdk-go/blob/master/types/operator.go#L39

**To Reproduce**

**Expected behavior**
Pass validation if user has 0x0 as DelegationApproverAddress

**Screenshots**

**OS details**

**Additional context**

shrimalmadhur:
Fixes #29 

### Solution
Also added a body.Close() fix

### DIFF
--- a/types/operator.go
+++ b/types/operator.go
@@ -63,7 +63,7 @@ func (o Operator) Validate() error {
 			fmt.Println("error closing metadata url body")
 		}
 	}(resp.Body)
-	
+
 	body, err := io.ReadAll(resp.Body)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


samlaf (approved):
Lgtm

PR #35, opened 2023-10-12
add batch key create utility

### LINKED ISSUE #30: Batch Key creation tool
Implement a tool which can create keys in batches for testing. Since this is for AVS testing, we are creating a utility in SDK. 

Note: This utility should be only used for testing

both for bls/ecdsa
- A private.keys.txt file with one private key hex in each line
- A folder which has all the encrypted json files
- A private.keys.loc.txt file with one filename for each private key
- A passwords.txt file with one password for each of those keys

**Additional context**

shrimalmadhur:
Fixes #30 

### Motivation
Create a new utility for batch key creation


shrimalmadhur:
any name suggestion? geth has a utility name `ethkey` - I out ours as `elkey`

jianoaix (dismissed):
Rubber stamp

utils/test_data/generated/password.txt
samlaf, line 1: was this a mistake or you really want to check this in?

utils/test_data/generated/private_key_hex.txt
samlaf, line 1: was this a mistake or you really want to check this in?

cmd/elkey/README.md
```diff
@@ -0,0 +1,19 @@
+## elkey
```
samlaf, line 1: is el here for eigenlayer or execution layer?
Can't use el for eigenlayer like this, since its extremely confusing.
We've been told this by EF as well afaik.
maybe `egnkey`?

utils/test_data/generated/password.txt
samlaf: Didn't we say we were going to generate and stores keys in `$HOME/.eigenlayer/keys` or something?

utils/batch_key.go
```diff
@@ -0,0 +1,97 @@
...
+
+import (
+	"bufio"
+	"fmt"
+	"os"
+	"path/filepath"
+)
+
+const (
+	// TODO: move to a common constatns file which
```
samlaf, line 11: constants**

utils/batch_key.go
```diff
@@ -0,0 +1,97 @@
...
+	// can be used by utils and cmd
+	DefaultKeyFolder  = "keys"
```
samlaf, line 13: shouldn't this be `$HOME/.eigenlayer/keys`?

utils/types.go
```diff
@@ -0,0 +1,7 @@
+package utils
+
+type BatchKey struct {
```
samlaf, line 3: Why do we need a separate struct for this? Why can't `CreateBatchKeys` just return a list of normal `Keys` (or whatever the other struct is called)?

cmd/elkey/main.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"fmt"
+	"os"
+
+	"github.com/urfave/cli/v2"
+)
+
+func main() {
+	app := cli.NewApp()
+	app.Name = "Eigenlayer batch key manager"
```
samlaf, line 12: "batch key" sounds weird to me. Maybe "batch keys" or "key batch"?

cmd/elkey/main.go
```diff
@@ -0,0 +1,25 @@
...
+	app.Commands = []*cli.Command{
+		commandGenerate,
+	}
+
+	if err := app.Run(os.Args); err != nil {
+		_, err := fmt.Fprintln(os.Stderr, err)
+		if err != nil {
+			return
+		}
+		os.Exit(1)
```
samlaf, line 22: Shouldn't this be reversed? You want to `os.Exit(1)` when there's an error and os.Exit(0) (aka return) otherwise

utils/batch_key.go
```diff
@@ -0,0 +1,97 @@
...
+// folder: folder where the keys are stored, relative to the current directory
+func ReadBatchKeys(folder string, isECDSA bool) ([]BatchKey, error) {
+	var batchKey []BatchKey
+	absFolder, err := filepath.Abs(folder)
+	if err != nil {
+		return nil, err
+	}
+
+	// read the private key file
+	privateKeyFile, err := os.Open(absFolder + "/" + PrivateKeyHexFile)
```
samlaf, line 29: this path won't work on windows. are we fine with that? otherwise use os.path or whatever golang stdlib for paths

utils/batch_key.go
```diff
@@ -0,0 +1,97 @@
...
+	defer func(privateKeyFile *os.File) {
+		err := privateKeyFile.Close()
+		if err != nil {
+			_ = fmt.Errorf("error closing the file: %s", err)
+			return
+		}
+	}(privateKeyFile)
+
+	// read the password file
+	passwordFile, err := os.Open(absFolder + "/" + PasswordFile)
```
samlaf, line 43: use proper path lib

cmd/elkey/generate.go
```diff
@@ -0,0 +1,196 @@
...
+)
+
+var (
+	KeyTypeFlag = &cli.StringFlag{
+		Name:     "key-type",
+		Usage:    "key type to create (ecdsa or bls)",
+		Required: true,
+	}
+	NumKeysFlag = &cli.IntFlag{
+		Name:     "num-keys",
```
samlaf, line 32: should this be `max-num-keys` instead? What if `private_key_hex.txt` has fewer than `num-keys` lines?

cmd/elkey/generate.go
```diff
@@ -0,0 +1,196 @@
...
+		Usage:    "number of keys to create",
+		Required: true,
+	}
+	OutputDirFlag = &cli.StringFlag{
```
samlaf, line 36: I would personally prefer if we added a `--private-keys-file` and `--passwords-file` flags instead of needing to cd into the correct directory. Makes it more flexible and easier to use in bash scripts.

utils/test_data/generated/password.txt
shrimalmadhur: for testing

utils/test_data/generated/private_key_hex.txt
shrimalmadhur: for testing

cmd/elkey/generate.go
shrimalmadhur: `num-keys` here is how many key we want to generate - we create all the files here

utils/types.go
shrimalmadhur: What other struct. we are returning array of `BatchKey` so that we can read each keys password and hex and encrypted file.

cmd/elkey/main.go
shrimalmadhur: `batch keys` is good. I will update

cmd/elkey/README.md
shrimalmadhur: `egnkey` it is

utils/batch_key.go
shrimalmadhur: I guess. we can use a stdlib. there's a filepath.Clean utility I can use.

cmd/elkey/generate.go
shrimalmadhur: this is just the output directory where keys will get created - check the `test_data/generated` folder for the directory structure

cmd/elkey/main.go
shrimalmadhur: this whole thing is in error condition
```
_, err := fmt.Fprintln(os.Stderr, err)
		if err != nil {
			return
		}
		os.Exit(1)
```
but maybe there is a better way to handle.

cmd/elkey/generate.go
samlaf: But it reads the private keys from that hex file right?

cmd/elkey/generate.go
samlaf: I know. I’m saying you should add 2 extra flags to also make the input file location configurable

cmd/elkey/main.go
samlaf: Ah right. Then remove the if err return. You want to always exit with 1 since the app failed.

shrimalmadhur: yea removed. and just logged error

cmd/elkey/generate.go
shrimalmadhur: no no - this utility just create `num-keys` - new keys

samlaf: Ahh I see. I was misunderstanding how this is used.

cmd/egnkey/README.md
```diff
@@ -0,0 +1,19 @@
+## egnkey
+
+### How to build
+Navigate to [egnkey](../egnkey/) directory and run
```
samlaf, line 4: We should have instructions for how to install via go install instead. This way we can use this binary anywhere

cmd/egnkey/README.md
```diff
@@ -0,0 +1,19 @@
...
+```bash
+   go build
+```
+
+### How to run
+
+To create in a random folder
+```bash
+   ./egnkey generate --key-type ecdsa --num-keys <num_key>
```
samlaf, line 13: This is ugly. Egnkey should be installed in GOBIN so that it can be run as any other global bin program

samlaf: Actually this makes me think. Why isn’t this just a subcommand of Eigenlayer cli instead?

shrimalmadhur: coz - this is not for operators - this is for AVS devs to test. remember I was telling you how go-ethereum also have tools - I agree this installation is not ideal. I just wanted to create something which can unblock DA key creation. can add it later. I also need to add another funding utility

cmd/egnkey/README.md
shrimalmadhur: yea. agreed. will add it in next PR. I need to add a funding utility too for DA

cmd/elkey/generate.go
shrimalmadhur: there's no input file in this utility.  it creates password and private key file

cmd/egnkey/README.md
shrimalmadhur: added `go install` instructions

cmd/egnkey/README.md
samlaf: gotttcha. ok I'll accept this PR as soon as readme is updated.

cmd/elkey/generate.go
shrimalmadhur: So it's easier for eigenda to read. they wanted both due to existing state of tests. logically you are right. we are just generating it so that it can be used both ways easily.

shrimalmadhur: discussed offline. this is due to DA needs

samlaf (approved):
LGTM

PR #47, opened 2023-10-19
fixed registration logic

samlaf:
bls pubkey compendium registration logic was changed to also require signing over the bls pubkey compendium's address, so updated the offchain logic to also include this.

### DIFF
--- a/chainio/elcontracts/writer.go
+++ b/chainio/elcontracts/writer.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"math/big"
 
-	"github.com/ethereum/go-ethereum/common"
 	gethcommon "github.com/ethereum/go-ethereum/common"
 	gethtypes "github.com/ethereum/go-ethereum/core/types"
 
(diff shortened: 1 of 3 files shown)


chainio/elcontracts/writer.go
```diff
@@ -176,14 +178,15 @@ func (w *ELChainWriter) RegisterBLSPublicKey(
 	ctx context.Context,
 	blsKeyPair *bls.KeyPair,
 	operator types.Operator,
+	blsPubkeyCompendiumAddr common.Address,
```
shrimalmadhur, line 181: we don't need to pass it , we are already initializing in ELContractsClient - you can just call `GetBLSPublicKeyCompendiumContractAddress` - https://github.com/Layr-Labs/eigensdk-go/blob/master/chainio/clients/el_contracts_client.go#L68

samlaf: addressed in [aa70c37](https://github.com/Layr-Labs/eigensdk-go/pull/47/commits/aa70c37f9a6305568dbf4b90a56a44ad2bdcca7b)

shrimalmadhur (approved):

PR #48, opened 2023-10-19
add eigenpod binding

gpsanant:
adds the eigenpod binding


shrimalmadhur:
@gpsanant can you add this in https://github.com/Layr-Labs/eigensdk-go/blob/master/contracts/generate-bindings.sh ? - otherwise if someone runs generate, it will delete the bindings

shrimalmadhur (approved):

PR #49, opened 2023-10-21
git pull eigenlayer-contracts to master

samlaf:
only 1 big difference is that updateStakes() on the stakeRegistry no longer pushes an update to the slasher (no slasher for M2) so we got rid of the unneeded operatorId and prevElement params to the call.

Btw I think we always have small diffs in the bindings generated because we're on different versions of abigen. I'm on `1.13.2-stable`.


shrimalmadhur:
> Btw I think we always have small diffs in the bindings generated because we're on different versions of abigen. I'm on `1.13.2-stable`.

this is interesting. We should have a consistent way of abigen. one idea is to do it in docker where you install the version you specify, it mounts and generate in a folder and copies back. This will not have impact on local version.

shrimalmadhur (approved):

PR #54, opened 2023-10-24
now fetching operatorId dynamically in economicCollector

samlaf:
### Motivation
eigenDA team was running into errors because the economicCollector was fetching operatorId from the eigenda registry coordinator in its constructor, which is initialized before the node starts. When the operator is registered at node start, it only registers in the start() function, which happens after, so the economicCollector never gets updated with the operatorId and hence keeps erroring on the collect call.

### Solution
This PR moves fetching the operatorId into the collector. This adds 1 round trip latency to the node on every collect (every ~15sec), but at least this will fix this problem. Another solution might be to cache the operatorId after one of the collect calls gets the operatorId, but this *could* break if we later allow the operatorId to change.

### DIFF
--- a/metrics/collectors/economic/economic.go
+++ b/metrics/collectors/economic/economic.go
@@ -28,7 +28,6 @@ type Collector struct {
 	logger            logging.Logger
 	// params to query the metrics for
 	operatorAddr common.Address
-	operatorId   types.OperatorId
 	quorumNames  map[types.QuorumNum]string
 	// metrics
 	// TODO(samlaf): I feel like eigenlayer-core metrics like slashingStatus and delegatedShares, which are not avs specific,
@@ -64,16 +63,11 @@ func NewCollector(
 	avsName string, logger logging.Logger,
 	operatorAddr common.Address, quorumNames map[types.QuorumNum]string,
 ) *Collector {
-	operatorId, err := avsRegistryReader.GetOperatorId(context.Background(), operatorAddr)
-	if err != nil {
-		logger.Error("Failed to get operator id", "err", err)
-	}
 	return &Collector{
 		elReader:          elReader,
 		avsRegistryReader: avsRegistryReader,
 		logger:            logger,
 		operatorAddr:      operatorAddr,
-		operatorId:        operatorId,
 		quorumNames:       quorumNames,
 		slashingStatus: prometheus.NewDesc(
 			types.EigenPromNamespace+"_slashing_status",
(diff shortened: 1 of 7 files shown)


metrics/collectors/economic/economic.go
```diff
@@ -135,20 +129,28 @@ func (ec *Collector) Collect(ch chan<- prometheus.Metric) {
 	// TODO(samlaf): implement this. probably have to call the BLSOperatorStateRetriever contract?
 	// probably should start using the avsregistry service instead of chainio clients so that we can
 	// swap out backend for a subgraph eventually
-	quorumNums, blsOperatorStateRetrieverOperator, err := ec.avsRegistryReader.GetOperatorsStakeInQuorumsOfOperatorAtCurrentBlock(context.Background(), ec.operatorId)
+	operatorId, err := ec.avsRegistryReader.GetOperatorId(context.Background(), ec.operatorAddr)
 	if err != nil {
-		ec.logger.Error("Failed to get operator stake", "err", err)
+		ec.logger.Error("Failed to get operator id", "err", err)
```
shrimalmadhur, line 134: shouldn't you return here if there's an error?

samlaf: Not returning, but shouldn't be doing the rest of the logic. Fixed: [9ab28a8](https://github.com/Layr-Labs/eigensdk-go/pull/54/commits/9ab28a8a5753442c03ae3466598c6b1168384328)

metrics/collectors/economic/economic.go
```diff
@@ -135,20 +129,29 @@ func (ec *Collector) Collect(ch chan<- prometheus.Metric) {
...
-						ec.registeredStake, prometheus.GaugeValue, stakeFloat64, strconv.Itoa(int(quorumNum)), ec.quorumNames[quorumNum],
-					)
+		quorumNums, blsOperatorStateRetrieverOperator, err := ec.avsRegistryReader.GetOperatorsStakeInQuorumsOfOperatorAtCurrentBlock(context.Background(), operatorId)
+		if err != nil {
+			ec.logger.Error("Failed to get operator stake", "err", err)
+		} else {
+			for quorumIdx, quorumNum := range quorumNums {
+				// TODO: this is stupid.. when AVSs scale to have 5K operators we'll be running through a bunch of operators
+				// we should instead just call registryCoordinator.getQuorumBitmapIndicesByOperatorIdsAtBlockNumber
+				// and stakeRegistry.getStakeForOperatorIdForQuorumAtBlockNumber directly
```
jianoaix, line 147: Why don't we do these two per-operator fetches?

jianoaix (commented):
It looks reasonable to cache the operatorID, as it's something extremely rare (and difficult) to change?

metrics/collectors/economic/economic.go
samlaf: Because I was lazy at the time. Fixed: [858ae16](https://github.com/Layr-Labs/eigensdk-go/pull/54/commits/858ae16cda0dcdcf98ec391c40af565a00c39edd)

metrics/collectors/economic/economic.go
```diff
@@ -110,6 +108,17 @@ func (ec *Collector) Describe(ch chan<- *prometheus.Desc) {
 	// ch <- ec.delegatedShares
 }
 
+func (ec *Collector) cacheOperatorIdIfNotCached() error {
```
shrimalmadhur, line 111: `cacheOperatorIdIfNotCached` - it's a weird name. don't `cache` already mean if some key is not present it would cache it?

samlaf: Not in this case because we only cache the very first time this is called. If it's already cached we don't update the cache.

samlaf: maybe `maybeCacheOperator`? or `cacheOperatorIfNeeded`? feel free to find another name, but I don't think `cacheOperatorId` works

shrimalmadhur (dismissed):

metrics/collectors/economic/economic.go
samlaf: Renamed to `initOperatorId` as we discussed on slack: [fa42170](https://github.com/Layr-Labs/eigensdk-go/pull/54/commits/fa42170176c0d4c501ba4fe192d8704e20ccad9f)

jianoaix: IMO just something like `getOperatorID()` will be good. The cache is internal detail/optimization that the caller doesn't need to care about.

jianoaix (approved):

PR #60, opened 2023-10-25
add validation to generates 32 bytes keys

shrimalmadhur:
Fixes #58

### Motivation
`crypto.GenerateKey()` randomly generate 31 bytes private keys.  go-ethereum doesn't like that so we are adding validations for that.

PR #61, opened 2023-10-25
[EgnKey] Generate ECDSA and BLS keys with both-key-type

siddimore:
Fixes # .

### Motivation
1. Create ECDSA and BLS keys using both key-type 
2. Prefix UUID folder name with ecdsa- and bls- to know the dir type

### Solution
1. Move some methods to common
2. Add key-type both

### DIFF
--- a/cmd/egnkey/generate.go
+++ b/cmd/egnkey/generate.go
@@ -222,6 +222,7 @@ func generateECDSAKeys(numKeys int, path string, passwordFile, privateKeyFile *o
 		key, err := crypto.GenerateKey()
 		privateKeyHex := hex.EncodeToString(key.D.Bytes())
 
+		// Low tech fix for creating valid keys, probably better is validation code
 		// Checking if the length of privateKeyHex is 32 bytes (64 characters)
 		for {
 			if len(privateKeyHex) != 64 {
--- a/cmd/egnkey/generate.go
+++ b/cmd/egnkey/generate.go
@@ -222,8 +222,8 @@ func generateECDSAKeys(numKeys int, path string, passwordFile, privateKeyFile *o
 		key, err := crypto.GenerateKey()
 		privateKeyHex := hex.EncodeToString(key.D.Bytes())
 
-		// Low tech fix for creating valid keys, probably better is validation code
-		// Checking if the length of privateKeyHex is 32 bytes (64 characters)
+		// Check if the length of privateKeyHex is 32 bytes (64 characters)
+		// Validation in Library will handle this
 		lenPrivateKey := len(privateKeyHex)
 		if lenPrivateKey != 64 {
 			fmt.Printf("Private key Ignore: %s %d\n", privateKeyHex, lenPrivateKey)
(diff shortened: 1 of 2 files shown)


cmd/egnkey/generate.go
```diff
@@ -153,6 +220,19 @@ func generateBlsKeys(numKeys int, path string, passwordFile, privateKeyFile *os.
...
+		privateKeyHex := hex.EncodeToString(key.D.Bytes())
+
+		// Low tech fix for creating valid keys, probably better is validation code
+		// Checking if the length of privateKeyHex is 32 bytes (64 characters)
+		for {
+			if len(privateKeyHex) != 64 {
+				key, err = crypto.GenerateKey()
+				privateKeyHex = hex.EncodeToString(key.D.Bytes())
+			} else {
+				break
```
shrimalmadhur, line 232: I don't think we need a for loop to check this - we should just ignore the key and let the upper loop continue - I actually already have a PR for this - https://github.com/Layr-Labs/eigensdk-go/pull/60

cmd/egnkey/generate.go
```diff
@@ -153,6 +220,16 @@ func generateBlsKeys(numKeys int, path string, passwordFile, privateKeyFile *os.
...
 	for i := 0; i < numKeys; i++ {
 		key, err := crypto.GenerateKey()
+		privateKeyHex := hex.EncodeToString(key.D.Bytes())
+
+		// Low tech fix for creating valid keys, probably better is validation code
+		// Checking if the length of privateKeyHex is 32 bytes (64 characters)
+		lenPrivateKey := len(privateKeyHex)
+		if lenPrivateKey != 64 {
+			fmt.Printf("Private key Ignore: %s %d\n", privateKeyHex, lenPrivateKey)
+			continue
```
shrimalmadhur, line 230: we will need to do `i--` otherwise it will generate less keys

shrimalmadhur:
can you also add this log message - then I can abandon my PR - https://github.com/Layr-Labs/eigensdk-go/pull/60/files#diff-530ece94ffa448cc7378e79ed63557d2f0f454ff2e904ee1411c197a4a3820e3R151 

can add these in both ecdsa, bls

shrimalmadhur:
Also please update the readme for the usage of `both` type and explain what it does.

cmd/egnkey/generate.go
```diff
@@ -153,6 +220,16 @@ func generateBlsKeys(numKeys int, path string, passwordFile, privateKeyFile *os.
 func generateECDSAKeys(numKeys int, path string, passwordFile, privateKeyFile *os.File) error {
 	for i := 0; i < numKeys; i++ {
 		key, err := crypto.GenerateKey()
+		privateKeyHex := hex.EncodeToString(key.D.Bytes())
+
+		// Check if the length of privateKeyHex is 32 bytes (64 characters)
+		// Validation in Library will handle this
```
shrimalmadhur, line 226: `Validation in Library will handle this` - what does this comment means?

shrimalmadhur (approved):

PR #63, opened 2023-10-27
update metadata to only allow png images

shrimalmadhur:
Fixes # .

### Motivation

We only want to allow `png` image upload to reduce surface are of image validation. IF we need other image format like jpg/jpeg, we can add that later.

### Solution
- Update validation
- Add tests

### DIFF
--- a/types/operator_metadata.go
+++ b/types/operator_metadata.go
@@ -111,7 +111,10 @@ func isImageURL(urlString string) bool {
 	extension := filepath.Ext(path)
 
 	// List of common image file extensions
-	imageExtensions := []string{".jpg", ".jpeg", ".png", ".gif", ".bmp", ".svg", ".webp"}
+	// Only support PNG for now to reduce surface are of image validation
+	// We do NOT want to support formats like SVG since they can be used for javascript injection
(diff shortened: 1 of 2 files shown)


types/operator_metadata.go
```diff
@@ -111,7 +111,10 @@ func isImageURL(urlString string) bool {
 	extension := filepath.Ext(path)
 
 	// List of common image file extensions
-	imageExtensions := []string{".jpg", ".jpeg", ".png", ".gif", ".bmp", ".svg", ".webp"}
+	// Only support PNG for now to reduce surface are of image validation
```
samlaf, line 114: are -> area

types/operator_metadata.go
```diff
@@ -111,7 +111,10 @@ func isImageURL(urlString string) bool {
...
+	// We do NOT want to support formats like SVG since they can be used for javascript injection
+	// If we get pushback on only supporting png, we can  jpg, jpeg, gif, etc. later
```
samlaf, line 116: we can ?

samlaf (approved):
LGTM. 2 small typos in comments

PR #68, opened 2023-10-31
implemented store command for ecdsa

### LINKED ISSUE #67: no way for egnkey to convert my hex private key to web3 secret storage format
Currently egnkey binary only allows me to generate a bunch of keys.
My use case is that I have my private key, which I can export from metamask in hex, and I want to be able to store it in a web3 secret storage file to comply with incredible-squaring operator requirements (see [config file](https://github.com/Layr-Labs/incredible-squaring-avs/blob/ea4b6551ac5e26ad5cf24e43ca6e900003e90509/config-files/operator.anvil.yaml#L23))

samlaf:
Fixes #67  .

@shrimalmadhur let me know what you think. Couldn't figure out how you generated these files, so I went ahead and added this functionality. Did you have another (simpler?) way of doing this already?

### DIFF
--- a/cmd/egnkey/main.go
+++ b/cmd/egnkey/main.go
@@ -13,6 +13,7 @@ func main() {
 	app.Description = "Eigenlayer batch keys manager"
 	app.Commands = []*cli.Command{
 		commandGenerate,
+		commandStore,
 	}
 
 	app.Usage = "Used to manage batch keys for testing"
--- a/cmd/egnkey/store.go
+++ b/cmd/egnkey/store.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"github.com/Layr-Labs/eigensdk-go/crypto/ecdsa"
-	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/urfave/cli/v2"
 )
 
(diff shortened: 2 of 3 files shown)


shrimalmadhur:
@samlaf https://github.com/NethermindEth/eigenlayer/blob/develop/cli/operator/keys/import.go this is mostly on eigenlayer-cli side where we can do that one by one. Maybe for egnkey, since its more focused for AVS devs, we can have bulk import? like provide a file with private keys hexes, one key per line and it can generate password and json files.

samlaf:
> @samlaf https://github.com/NethermindEth/eigenlayer/blob/develop/cli/operator/keys/import.go this is mostly on eigenlayer-cli side where we can do that one by one since that's for operator. Maybe for egnkey, since its more focused for AVS devs, we can have bulk import? like provide a file with private keys hexes, one key per line and it can generate password and json files.
> 
> But I do think we can have the main logic of saving the file in `crypto/utils` and `egnkey` and `eigenlayer-cli` both can use it.

added new fct WriteKeyFromHex, so that you can reuse it in the cli

shrimalmadhur (approved):

PR #73, opened 2023-11-03
refactor chainio clients

### LINKED ISSUE #26: Refactor ChainIO clients structure
Current chainio clients are overly complex and are pretty frustrating to work with (see for eg the chainio/constructor flow for constructing all clients).

The current clients hierarchy looks like
image
but we'd rather it look like
image

```[tasklist]
### Tasks
- [ ] https://github.com/Layr-Labs/eigensdk-go/issues/25
```

samlaf:
Fixes #26  .

Managed to get rid of the elChainClient struct by forking abigen to generate interfaces for the structs it generates, and using those interfaces directly in the elWriter/elReader/elSubscriber structs. There's still a bit of clutter (see the few TODOs I left). I think for example we should change the interface of the writers/readers/subscribers to take the geth Call/TxOpts directly, instead of taking a ctx and wrapping it manually in a Call/TxOpts. Also should be reusing the bindings struct to have nicer constructors. I left the elContractBindings struct there for this purpose, since I think we should repurpose it as a constructor (maybe move it to chainio/constructors actually..)

@shrimalmadhur @NimaVaziri please review these changes. If you like them, I'll do the same for the AVS clients.


contracts/bindings/EigenPod/binding.go
```diff
@@ -1,2305 +0,0 @@
-// Code generated - DO NOT EDIT.
```
shrimalmadhur, line 1: Why is this file getting deleted?

chainio/constructor/constructor.go
```diff
@@ -115,11 +112,13 @@ func BuildClients(config Config, logger logging.Logger) (*Clients, error) {
 
 }
 
-func (config *Config) buildElContractsChainClient(
-	logger logging.Logger,
+func (config *Config) buildElClients(
```
shrimalmadhur, line 115: Can we export this method? - eigenlayer-cli just cares about this and doesn't care about avs registry stuff. so it would be better to use this method.

chainio/elcontracts/reader.go
```diff
@@ -60,29 +66,42 @@ type ELReader interface {
...
+	strategyManager         strategymanager.ContractStrategyManagerCalls
+	blsPubkeyCompendium     blspkcompendium.ContractBLSPublicKeyCompendiumCalls
+	blsPubKeyCompendiumAddr common.Address
+	ethClient               eth.EthClient
 }
 
 // forces EthReader to implement the chainio.Reader interface
 var _ ELReader = (*ELChainReader)(nil)
 
 func NewELChainReader(
```
shrimalmadhur, line 81: This constructor has become very big. as in what if someone just want to initialize reader and not writer? 
two ways we can handle
1) We make this method private and force users to use constructor
2) We have another method which just accepts few contract addresses and logger, ethclient and just give reader back

I prefer 2 since it's flexible. 


same for writer/subscriber

contracts/bindings/EigenPod/binding.go
samlaf: Not sure… generating-bindings fails on this file for some weird reason which I haven’t debugged yet.

chainio/elcontracts/reader.go
samlaf: Don’t get your question. This is just the reader constructor. Why are you asking what happens if they want to not build Writer? Was this comment meant to be about constructor.go instead of reader.go?

chainio/constructor/constructor.go
samlaf: Yes absolutely we need to clean this constructor file to have smaller chunk initializations. Will be useful also for egnaddrs. Need to think how to organize all of this including building the bindings in bindings.go (could be moved here as a helper method)

chainio/elcontracts/reader.go
samlaf: Your 2 sounds like what I was proposing in https://github.com/Layr-Labs/eigensdk-go/pull/11

contracts/bindings/EigenPod/binding.go
samlaf: Fixed in [c5284ae](https://github.com/Layr-Labs/eigensdk-go/pull/73/commits/c5284aee3638d34d4925a97a1e39ff27236ba73f)
Super weird bug though... EigenPod needs to be abigen'd right before EigenPodManager otherwise it fails for some unknown reason.. I left comment, not going to spend time debugging this.

chainio/elcontracts/reader.go
samlaf: Added EL constructors, for eg https://github.com/Layr-Labs/eigensdk-go/pull/73/commits/3081f8a420ab2cb4608c875b187ff4977918a2e7#diff-78f2b53998897f8fe40be2a3043dec1f342d78f98c9ef5b5abf867a8160fe938R102

chainio/constructor/constructor.go
samlaf: Added EL constructors, for eg https://github.com/Layr-Labs/eigensdk-go/pull/73/commits/3081f8a420ab2cb4608c875b187ff4977918a2e7#diff-78f2b53998897f8fe40be2a3043dec1f342d78f98c9ef5b5abf867a8160fe938R102

samlaf:
@shrimalmadhur need new review. This is almost ready to go. One thing I'm thinking of doing though is changing the interfaces for all our el/avs readers/writers/subscribers, and have them expose the same interface as the bindings themselves (eg. take a callOpts or transactOpts directly instead of a ctx). Thoughts?

Also did you change the settings of the linter at any point? I didn't change anything with the metrics examples afaik and not it's complaining that EigenMetrics is an unknown identifier (I think because EigenMetrics is in metrics pkg whereas the example test is in metrics_test pkg.. but unsure)

samlaf:
Ok so main question left to complete this refactor @shrimalmadhur @NimaVaziri is what interface we expose for the el/avs reader/writer/subscriber clients. Right now we only expose some ad-hoc methods and add them as we get new needs. For eg 
image

2 remaining questions for me are:
- as I mentioned in the previous comment, I think we should get rid of the ctx arguments and replace them directly with the geth.bind TransactOpts/CallOpts/FilterOpts which wrap a context with extra arguments and give more flexibility to the caller for how they wish the tx to be sent.
- do we want to keep these APIs minimal, or should we just expose ALL of the underlying binding methods (all the ocntract methods basically). I think there is a problem with doing this in that if we just use struct embeddings there might be method conflicts b/c all the different contracts are initializable for example so all have the same init() method which go doesn't like... but there might be a way (?)

chainio/clients/elcontracts/reader.go
```diff
@@ -60,29 +67,63 @@ type ELReader interface {
...
+		elContractBindings.DelegationManager,
+		elContractBindings.StrategyManager,
+		elContractBindings.BlsPubkeyCompendium,
+		blsPubKeyCompendiumAddr,
+		logger,
+		ethClient,
+	), nil
 }
 
+// TODO(samlaf): should we just pass the CallOpts directly as argument instead of the context?
```
shrimalmadhur, line 123: the reason we don't want it - what if over time abigen upgrades and deprecate this CallOpts for something else. Our interface will break - in current we hide all the dependencies interface and objects so that probability of breaking changes is low

chainio/clients/elcontracts/reader.go
```diff
@@ -131,24 +172,50 @@ func (r *ELChainReader) GetOperatorDetails(ctx context.Context, operator types.O
...
+		r.logger.Error("Failed to fetch strategy contract", "err", err)
+		return nil, common.Address{}, err
+	}
+	underlyingTokenAddr, err := contractStrategy.UnderlyingToken(&bind.CallOpts{})
+	if err != nil {
+		r.logger.Error("Failed to fetch token contract", "err", err)
+		return nil, common.Address{}, err
+	}
+	return contractStrategy, underlyingTokenAddr, nil
 }
```
shrimalmadhur, line 191: nit: add a // comment tag here too.

chainio/clients/elcontracts/writer.go
```diff
@@ -41,37 +44,93 @@ type ELWriter interface {
...
+		blsPubKeyCompendiumAddr,
+		elChainReader,
+		ethClient,
+		signer,
+		logger,
+		eigenMetrics,
+	), nil
+}
+
+// TODO(madhur): do we really want to wait for txreceipts like this in these functions?
```
shrimalmadhur, line 124: let's keep it right now. This will change once we have a better tx manager design implemented in  #75

chainio/clients/elcontracts/reader.go
```diff
@@ -60,29 +67,63 @@ type ELReader interface {
...
+	strategyManager         strategymanager.ContractStrategyManagerCalls
+	blsPubkeyCompendium     blspkcompendium.ContractBLSPublicKeyCompendiumCalls
+	blsPubKeyCompendiumAddr common.Address
+	ethClient               eth.EthClient
 }
 
 // forces EthReader to implement the chainio.Reader interface
 var _ ELReader = (*ELChainReader)(nil)
 
 func NewELChainReader(
```
shrimalmadhur, line 82: Do we want to make this private if we want to enforce them to use `BuildELChainReader`?

chainio/clients/elcontracts/writer.go
```diff
@@ -41,37 +44,93 @@ type ELWriter interface {
...
+	blsPubkeyCompendiumAddr gethcommon.Address
+	elChainReader           ELReader
+	ethClient               eth.EthClient
+	signer                  signer.Signer
+	logger                  logging.Logger
 }
 
 var _ ELWriter = (*ELChainWriter)(nil)
 
 func NewELChainWriter(
```
shrimalmadhur, line 61: Same for this. do we want to make this private?

contracts/abigen-with-interfaces.Dockerfile
```diff
@@ -0,0 +1,17 @@
+# use the below command to build this docker image locally
```
shrimalmadhur, line 1: I am not against this idea but who's going to maintain your branch with abigen updates?  like what if there are fixes and upgrades to abigen which we want to incorporate?

chainio/utils/bindings.go
```diff
@@ -16,20 +18,26 @@ import (
...
+type EigenlayerContractBindings struct {
+	SlasherAddr             gethcommon.Address
+	StrategyManagerAddr     gethcommon.Address
+	DelegationManagerAddr   gethcommon.Address
+	BlspubkeyCompendiumAddr gethcommon.Address
+	Slasher                 *slasher.ContractSlasher
+	DelegationManager       *delegationmanager.ContractDelegationManager
+	StrategyManager         *strategymanager.ContractStrategyManager
+	// TODO: should this really be here? or moved to avs bindings
+	BlsPubkeyCompendium *blspubkeycompendium.ContractBLSPublicKeyCompendium
```
shrimalmadhur, line 32: I am very confused with this - seems like we deploy it with AVS today (with EigenDA deployment). how is it going to work when we have multiple AVS?

chainio/clients/builder.go
```diff
@@ -45,7 +44,7 @@ type Clients struct {
 	PrometheusRegistry     *prometheus.Registry  // Used if avs teams need to register avs-specific metrics
 }
 
-func BuildClients(config Config, logger logging.Logger) (*Clients, error) {
+func BuildAll(config BuildAllConfig, logger logging.Logger) (*Clients, error) {
```
shrimalmadhur, line 47: I think builder should also accept signer. this only supports privatekeyhex which is good for testing but what about bulding all with  the local keystore or remote signer etc? unless you are only intending this for tests?

chainio/clients/builder.go
```diff
@@ -18,7 +17,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
-type Config struct {
+type BuildAllConfig struct {
```
shrimalmadhur, line 20: maybe we should validate this config since it's a free flowing config and can have unintended panics if some configs are not filled?

chainio/clients/builder.go
```diff
@@ -45,7 +44,7 @@ type Clients struct {
 	PrometheusRegistry     *prometheus.Registry  // Used if avs teams need to register avs-specific metrics
```
shrimalmadhur, line 44: should we make this ETH clients private in these `Clients` struct? We should only expose the clients which we want them to use

chainio/clients/elcontracts/reader.go
samlaf: But geth has been around for much longer, so their interface is less likely to change frequently, whereas if we start doing this, its our interface that is going to be breaking all the time (oh this method we need more control on startBlock, let's add it to interface, etc.). Eventually we'll just end up with the geth interface on most functions.

chainio/clients/elcontracts/reader.go
samlaf: Done: [f67f711](https://github.com/Layr-Labs/eigensdk-go/pull/73/commits/f67f71154866941837b176889c0f592576ebadb0)

chainio/clients/elcontracts/writer.go
samlaf: Updated comment: [3d14cf3](https://github.com/Layr-Labs/eigensdk-go/pull/73/commits/3d14cf3a2662924a57d01e098c987bd33ce8efcf)

chainio/clients/elcontracts/reader.go
samlaf: We discussed offline. Let's keep it for now. It's used in the chainio/cleints/builder constructor. We can revisit the design of these clients (what we want private, what we want public, which constructors are required, which should we hide, etc.) in a later refactor.

chainio/clients/elcontracts/writer.go
samlaf: We discussed offline. Let's keep it for now. It's used in the chainio/cleints/builder constructor. We can revisit the design of these clients (what we want private, what we want public, which constructors are required, which should we hide, etc.) in a later refactor.

contracts/abigen-with-interfaces.Dockerfile
samlaf: I'll have to maintain it. that's just the nature of software. Pick your poison: either I maintain this or I have to maintain writing interfaces by hand and updating them much more frequently.

I have a PR open on geth to merge these changes: https://github.com/ethereum/go-ethereum/issues/28278

chainio/utils/bindings.go
samlaf: We originally were thinking to have a single one shared between all teams. But I think the current thinking is that every team will deploy their own. I'm in favor of this change because it gives more freedom to operators to use different keys for different avss, albeit costs a bit more gas.

chainio/clients/builder.go
samlaf: good call. fixed in [32ca000](https://github.com/Layr-Labs/eigensdk-go/pull/73/commits/32ca0003afdcd450f7121dc54889238907c19abc)

chainio/clients/builder.go
samlaf: They might need it to call json-rpc methods like blockNumber. This was your insight remember? :P

chainio/clients/builder.go
samlaf: I think panicing inside a constructor is fine, at least it'll be caught pretty early.

chainio/clients/elcontracts/reader.go
samlaf: We agreed to make the changes. Updated to new api in [6e2c821](https://github.com/Layr-Labs/eigensdk-go/pull/73/commits/6e2c8217737086857aab8f15824fed3ce0d4688d) and [d999091](https://github.com/Layr-Labs/eigensdk-go/pull/73/commits/d9990916f28f96e754bf7674cfbb620faa5f1f2c)

samlaf:
@shrimalmadhur all changes are in. Ready for final review!

chainio/gomock_reflect_1102840134/prog.go
```diff
@@ -0,0 +1,66 @@
+
```
shrimalmadhur, line 1: delete this file. How did it come in first place 🤔

chainio/utils/bindings.go
shrimalmadhur: so if every AVS has different BLS pubkey compendium then we need to rethinking combining BLS pubkey registration via EigenLayer CLI since the current assumption was that there's only one.

contracts/abigen-with-interfaces.Dockerfile
shrimalmadhur: as long as we are good with maintaining it - we are good.

chainio/gomock_reflect_1102840134/prog.go
samlaf: oops. deleted: [a40c6be](https://github.com/Layr-Labs/eigensdk-go/pull/73/commits/a40c6be184ca467a4e361736f7494e0f7d0c8d52)

chainio/clients/builder.go
shrimalmadhur: I mean the problem is more on - you initialize all the other things unnecessarily if the panic happens later. so we might want to make sure config is correct before we start initializing anything.

chainio/clients/builder.go
shrimalmadhur: hmm. I guess this is just a mixture of clients so it's fine.

chainio/clients/builder.go
shrimalmadhur: discussed offline.

samlaf: Added in [95482b2](https://github.com/Layr-Labs/eigensdk-go/pull/73/commits/95482b2c9f3eb0bb60971a2672c263add9848d53)

shrimalmadhur (approved):

PR #77, opened 2023-11-17
return better error for image extension mismatch

### LINKED ISSUE #76: Error message should tell user about allowed image type
**Describe the bug**

Right now we throw a generic error message when image is invalid - https://github.com/Layr-Labs/eigensdk-go/blob/master/types/operator_metadata.go#L56

We should throw specific error that only `.png` is allowed and other specific error

**To Reproduce**

**Expected behavior**

**Screenshots**

**OS details**

**Additional context**

shrimalmadhur:
Fixes #76  .

### Motivation
Right now we don't have specific message for image extension error

### Solution
- Return proper error
- Dependency upgrades


types/operator_metadata.go
```diff
@@ -52,8 +52,8 @@ func (om *OperatorMetadata) Validate() error {
 		return errors.New("logo is required")
 	}
 
-	if !isImageURL(om.Logo) {
-		return errors.New("logo must be a valid image url")
+	if err := isImageURL(om.Logo); err != nil {
+		return errors.New("logo must be a valid image url with err: " + err.Error())
```
samlaf, line 56: I would just return the err here instead of adding more stuff, since unclear if it actually adds anything, plus it could also be moved down to the error in `isImageURL`.

samlaf (approved):
minor comment. I'm fine with merging without that change though, so approving.

types/operator_metadata.go
shrimalmadhur: yes. fixed

samlaf (approved):

PR #78, opened 2023-11-18
Use singular nouns for target names in makefile

estensen:
`make test` feels more idiomatic than `make tests`

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -6,7 +6,7 @@ on:
       - master
   pull_request:
 
-jobs:  
+jobs:
   Test:
     name: Test
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


samlaf:
@estensen do you have a link for this where they state singular is more idiomatic? singular sounds weird to me.
`make bindings` and `make mocks` sound better than the singular to me.
`make tests` sounds weird, but so does `make test` haha, so not sure about this one.

estensen:
@samlaf sorry for leaving you hanging.

I don't have any documentation for that it's more idiomatic, it's only based on my own opinions, so feel free to close. But I do think it's a good idea to be consistent between repos in the org. It makes it easier to work with.

samlaf:
Sorry closing this. Don't feel it's necessary at this point in time, and don't want to get into naming wars. We should have conventions that we follow everywhere, but until we have such a standard I don't want to impose anything.

PR #81, opened 2023-11-21
validate bls key file contents

shrimalmadhur:
Fixes # .

### Motivation
Right now if you put ecdsa key file in place of bls key, it creates a new pair and carries on

### Solution
We want to put validation so that correct file is used for creating correct key. I am not sure if there's a better way to validate bls key file content which we generate.

### DIFF
--- a/crypto/bls/attestation.go
+++ b/crypto/bls/attestation.go
@@ -249,7 +249,7 @@ func ReadPrivateKeyFromFile(path string, password string) (*KeyPair, error) {
 	// This is to prevent and make sure pubkey is present.
 	// ecdsa keys doesn't have that field
 	if encryptedBLSStruct.PubKey == "" {
-		return nil, fmt.Errorf("invalid bls key file. pubkey not found")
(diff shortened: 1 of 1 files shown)


crypto/bls/attestation.go
```diff
@@ -243,6 +243,15 @@ func ReadPrivateKeyFromFile(path string, password string) (*KeyPair, error) {
 		return nil, err
 	}
 
+	// Check if pubkey is present, if not return error
+	// There is an issue where if you specify ecdsa key file
+	// it still works and returns a keypair since the format of storage is same.
+	// This is to prevent and make sure pubkey is present.
+	// ecdsa keys doesn't have that field
+	if encryptedBLSStruct.PubKey == "" {
+		return nil, fmt.Errorf("invalid bls key file. pubkey not found")
```
samlaf, line 252: change to "pubkey field not found"

samlaf (dismissed):

samlaf (approved):

PR #86, opened 2023-12-06
#42: refactor types and utils

### LINKED ISSUE #42: Refactor types and utils
We need to refactor types and utils so that it's in some standard way and not create circular dependencies. 

- Yet to define

**Additional context**

RohanShrothrium:
Fixes #42 .

### Motivation
Clean up `utils` and `types` modules.

### Solution
Moved type validations out of utils. Typecasting is moved to the service itself.

### Open questions
Right now the `types` module imports the crypto module which should not be the case. Certain types defined in crypto modules are used across other modules of the same hierarchy. Should I accommodate this in the same PR?

### DIFF
--- a/types/common.go
+++ b/types/common.go
@@ -1,0 +1,34 @@
+package types
+
+import (
+	"errors"
+	"net/url"
+	"regexp"
+)
+
+func isValidEthereumAddress(address string) bool {
+	re := regexp.MustCompile("^0x[0-9a-fA-F]{40}$")
+	return re.MatchString(address)
+}
+
+func checkIfUrlIsValid(rawUrl string) error {
+	// Regular expression to validate URLs
+	urlPattern := regexp.MustCompile(`^(https?|ftp)://[^\s/$.?#].[^\s]*$`)
+
+	// Check if the URL matches the regular expression
+	if !urlPattern.MatchString(rawUrl) {
+		return errors.New("invalid url")
+	}
+
+	parsedURL, err := url.Parse(rawUrl)
+	if err != nil {
+		return err
+	}
+
+	// Check if the URL is valid
+	if parsedURL.Scheme != "" && parsedURL.Host != "" {
+		return nil
+	} else {
+		return errors.New("invalid url")
+	}
+}
--- a/types/operator.go
+++ b/types/operator.go
@@ -8,9 +8,9 @@ import (
 	"math/big"
 	"net/http"
 
-	"github.com/Layr-Labs/eigensdk-go/crypto/bls"
-	"github.com/Layr-Labs/eigensdk-go/utils"
 	"github.com/ethereum/go-ethereum/common"
+
+	"github.com/Layr-Labs/eigensdk-go/crypto/bls"
 )
 
 const (
@@ -32,15 +32,15 @@ type Operator struct {
 }
 
 func (o Operator) Validate() error {
-	if !utils.IsValidEthereumAddress(o.Address) {
+	if !isValidEthereumAddress(o.Address) {
 		return errors.New("invalid operator address")
 	}
 
-	if !utils.IsValidEthereumAddress(o.EarningsReceiverAddress) {
+	if !isValidEthereumAddress(o.EarningsReceiverAddress) {
 		return errors.New("invalid EarningsReceiverAddress address")
 	}
 
-	if o.DelegationApproverAddress != ZeroAddress && !utils.IsValidEthereumAddress(o.DelegationApproverAddress) {
+	if o.DelegationApproverAddress != ZeroAddress && !isValidEthereumAddress(o.DelegationApproverAddress) {
 		return fmt.Errorf(
 			"invalid DelegationApproverAddress address, it should be either %s or a valid non zero ethereum address",
 			ZeroAddress,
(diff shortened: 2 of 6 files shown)


utils/utils.go
```diff
@@ -2,58 +2,12 @@ package utils
...
 
-func ReadFile(path string) ([]byte, error) {
-	b, err := os.ReadFile(filepath.Clean(path))
-	if err != nil {
-		return nil, err
-	}
-	return b, nil
-}
-
-func ReadYamlConfig(path string, o interface{}) error {
```
shrimalmadhur, line 26: These methods are used in few of libraries we import eigensdk in. So let's not remove

utils/utils.go
```diff
@@ -2,58 +2,12 @@ package utils
...
-	"path/filepath"
-	"regexp"
-
 	gethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/crypto"
-	"gopkg.in/yaml.v3"
+	"math/big"
 )
 
-func ReadFile(path string) ([]byte, error) {
```
shrimalmadhur, line 18: Good call to remove it. I've no idea why I added this!

utils/utils.go
```diff
@@ -2,58 +2,12 @@ package utils
...
-
-	err = yaml.Unmarshal(b, o)
-	if err != nil {
-		log.Fatalf("unable to parse file with error %#v", err)
-	}
-
-	return nil
-}
-
-func ReadJsonConfig(path string, o interface{}) error {
```
shrimalmadhur, line 43: Please don't remove this!

utils/utils.go
```diff
@@ -71,8 +25,3 @@ func RoundUpDivideBig(a, b *big.Int) *big.Int {
 	res.Div(a, b)
 	return res
 }
-
-func IsValidEthereumAddress(address string) bool {
```
shrimalmadhur, line 75: This method is used by a library which imports eigensdk

shrimalmadhur (changes requested):

utils/utils.go
RohanShrothrium: Aaah makes sense. Also, can we move these functions to a different module like `eigenUtils` because intuitively `utils` are for internal use?

shrimalmadhur: ideally all internal stuff would've been in an `internal` package but we didn't follow that convention.

I am not sure what the ideal way of doing this. Are you suggesting we can have `utils` and `eigenUtils` both ? one for internal and one for external?

RohanShrothrium: Yep, this might cause breaking changes for the SDK users though.
Honestly, I don't see what these utils functions really abstract away for users. I believe the SDK should be used for its interfaces and the utils like reading config files should be handled by client apps.
What do you think?

utils/utils.go
samlaf: which library?

shrimalmadhur: @samlaf eigenlayer-cli - https://github.com/Layr-Labs/eigenlayer-cli/blob/master/pkg/operator/config/create.go#L195

utils/utils.go
shrimalmadhur: yea you are right. the only reason I added is that, multiple repository we have use this function and I just wanted to have it once. but I kind of agree it's on client to have these standard utils on their own. they can choose their own way to do this.

RohanShrothrium: Let me know the PoA for this PR @shrimalmadhur!

shrimalmadhur: Lets remove these function. I will make sure dependencies take care of it. @samlaf you good with that?

samlaf: Yep

shrimalmadhur: @RohanShrothrium feel free to make these changes

RohanShrothrium: @shrimalmadhur if you don't mind can you resolve all the comments which do not need any change? I am a bit confused as to what functions should be added back.

shrimalmadhur: done

RohanShrothrium: This function is removed as well in the initial commit ser!

samlaf: @shrimalmadhur is this good to be merged after we rebase on master?

shrimalmadhur: yep. @RohanShrothrium can you rebase and resolve conflicts? and then we can merge this

types/operator.go
```diff
@@ -32,15 +32,15 @@ type Operator struct {
 }
 
 func (o Operator) Validate() error {
-	if !utils.IsValidEthereumAddress(o.Address) {
+	if !isValidEthereumAddress(o.Address) {
```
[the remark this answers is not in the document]
shrimalmadhur: that's a good point. this PR is getting stale. might as well do in new PR. I will make an issue.

types/common.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+func isValidEthereumAddress(address string) bool {
+	re := regexp.MustCompile("^0x[0-9a-fA-F]{40}$")
+	return re.MatchString(address)
+}
+
+func checkIfUrlIsValid(rawUrl string) error {
+	// Regular expression to validate URLs
+	urlPattern := regexp.MustCompile(`^(https?|ftp)://[^\s/$.?#].[^\s]*$`)
```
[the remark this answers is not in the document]
shrimalmadhur: yea this is not perfect. as in we tailored it to our requirement and allowed certain endpoints, but definitely can add more.

PR #101, opened 2024-01-19
Make builder return txmgr

samlaf:
Fixes # .


samlaf:
After offline discussion with @shrimalmadhur, decided returning txMgr is not needed.
So created new PR https://github.com/Layr-Labs/eigensdk-go/pull/103 for the other feature in this PR.

PR #102, opened 2024-01-20
Fix isOperatorRegistered

MaanavKhaitan:
### Motivation

- `isOperatorRegistered()` was using the operator ID to determine if an operator is registered
- This was an issue because after an operator de-registers from an AVS, their operator info and ID is not removed from the registry coordinator so `isOperatorRegistered()` was still returning `true`

### Solution
- Fixes `isOperatorRegistered()` so it uses operator status instead of operator ID to determine if an operator is registered

### Testing
Tested this with the CLI in `incredible-squaring-avs` and verified that we're getting the correct value now for `registeredWithAvs`.

### DIFF
--- a/chainio/clients/avsregistry/reader.go
+++ b/chainio/clients/avsregistry/reader.go
@@ -360,13 +360,14 @@ func (r *AvsRegistryChainReader) IsOperatorRegistered(
 	opts *bind.CallOpts,
 	operatorAddress gethcommon.Address,
 ) (bool, error) {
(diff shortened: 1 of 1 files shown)


samlaf (approved):
Thank you sir, insta approve!
@shrimalmadhur the badge github action is still failing, prob because maanav is outside our org?

shrimalmadhur:
> Thank you sir, insta approve! @shrimalmadhur the badge github action is still failing, prob because maanav is outside our org?

yea. I have made it optional so as long as unit test passes, we are good to merge. haven't found a good way to make it work with folks outside the org

PR #104, opened 2024-01-25
Feat ecdsa sigs

samlaf:
We're currently developing ecdsa versions of everything:
- middleware registry contracts: https://github.com/Layr-Labs/eigenlayer-middleware/pull/123
   - @stevennevins and @ChaoticWalrus are working on cleaning these ecdsa registries properly. I made a very quick and dirty branch that I'm using for now https://github.com/Layr-Labs/eigenlayer-middleware/tree/ecdsa-with-history-poc/src/experimental
   - I reintroduced operator update histories onchain in this branch because it made it much easier to fork the bls aggregation service and inc-sq, but in Jeff's branch he removed histories. We're currently discussing what's the best way to remove the complexity in ecdsa contracts though, so this might change. See for eg https://eigenlabs.slack.com/archives/C05F0C8392L/p1706504508884089)
- incredible-squaring: https://github.com/Layr-Labs/incredible-squaring-avs/pull/20

Leaving this as a draft until I've finished incorporating these changes into the ecdsa version of inc-sq, but would still appreciate a review on the current structural decisions in this PR. Main things are:
- split avsregistry clients and aggregation service into ecdsa and bls ones
   - ideally we'd only have one aggregation service, possibly using generics, that could do both ecdsa and bls aggregation, but I just wanted something quick for now, so opted to duplicate a lot of the code. There are a lot of changes though, so don't think merging them would be that easy... we can prob get a junior dev to clean this up eventually (if we think its worth it)
- utils and types are getting ugly with the duplicate ecdsa and bls names, so please check those and suggest any better alternative for organizing these things


crypto/ecdsa/ecdsa.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+// SignMsg is a wrapper around geth's crypto.Sign function which adds 27 to the v value
+// see https://github.com/ethereum/go-ethereum/issues/28757#issuecomment-1874525854
+// and https://twitter.com/pcaversaccio/status/1671488928262529031
+// Currently all of our contracts expect the v value to be 27 or 28, which is why we add it here.
+// VerifySignature below thus subtracts 27 from the v value before verifying the signature.
+// TODO(samlaf): if and when we figure out some avs contracts use the 0/1 v scheme, or the eip-155 scheme,
+// then we should generalize this function to sign based on a scheme passed as argument.
+func SignMsg(msg []byte, privateKey *ecdsa.PrivateKey) ([]byte, error) {
```
samlaf, line 17: This is an issue that I'm unsure how to deal with, so arranged it like this for now, but if anyone has ideas on how to better deal with this, would be good to know.

shrimalmadhur:
Closing it because it is outdated

PR #107, opened 2024-01-31
add slog implementation

shrimalmadhur:
Fixes # .

### Motivation
Add basic slog implementation

### DIFF
--- a/chainio/clients/avsregistry/writer.go
+++ b/chainio/clients/avsregistry/writer.go
@@ -24,10 +24,11 @@ import (
 )
 
 type AvsRegistryWriter interface {
-	// TODO(samlaf): an operator that is already registered in a quorum can register with another quorum without passing signatures
-	//               perhaps we should add another sdk function for this purpose, that just takes in a quorumNumber and socket?
-	// RegisterOperatorInQuorumWithAVSRegistryCoordinator is used to register a single operator with the AVS's registry coordinator.
-	//  - operatorEcdsaPrivateKey is the operator's ecdsa private key (used to sign a message to register operator in eigenlayer's delegation manager)
+	// TODO(samlaf): an operator that is already registered in a quorum can register with another quorum without passing
+	// signatures perhaps we should add another sdk function for this purpose, that just takes in a quorumNumber and
+	// socket? RegisterOperatorInQuorumWithAVSRegistryCoordinator is used to register a single operator with the AVS's
+	// registry coordinator. - operatorEcdsaPrivateKey is the operator's ecdsa private key (used to sign a message to
+	// register operator in eigenlayer's delegation manager)
 	//  - operatorToAvsRegistrationSigSalt is a random salt used to prevent replay attacks
 	//  - operatorToAvsRegistrationSigExpiry is the expiry time of the signature
 	RegisterOperatorInQuorumWithAVSRegistryCoordinator(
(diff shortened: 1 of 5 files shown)


logging/slog_logger.go
```diff
@@ -0,0 +1,71 @@
...
+var _ Logger = (*SLogger)(nil)
+
+func NewSlogLogger(env LogLevel) Logger {
+	if env == Production {
+		logger := slog.New(slog.NewJSONHandler(os.Stdout, nil))
+		return &SLogger{
+			logger: logger,
+		}
+	} else if env == Development {
+		logger := slog.New(slog.NewTextHandler(os.Stdout, nil))
```
shrimalmadhur, line 22: I guess dev should also be json handler? since we see logs on grafana or anywhere

dmanc: Is TextHandler the same as logfmt?

shrimalmadhur: added

shrimalmadhur: @dmanc texthandler gives in this format 
`time=2022-11-08T15:28:26.000-05:00 level=INFO msg=hello count=3`

shrimalmadhur: I know we don't have structured logging in grafana yet (or we do? ) but I think structured logging needs json?

dmanc: Yeah texthandler is the same as logfmt it seems. Loki supports either:
- https://grafana.com/docs/loki/latest/send-data/promtail/stages/logfmt/
- https://grafana.com/docs/loki/latest/send-data/promtail/stages/json/

shrimalmadhur: that's cool. we can do either but should follow a standard everywhere.

chainio/clients/avsregistry/writer.go
samlaf, line 1: weird that these changes were included here. Did you checkout from a PR branch or something?

logging/slog_logger.go
```diff
@@ -0,0 +1,75 @@
...
+	if env == Production {
+		logHandler := slog.NewJSONHandler(os.Stdout, nil).WithAttrs([]slog.Attr{slog.String("env", string(env))})
+		logger := slog.New(logHandler)
+		return &SLogger{
+			logger: logger,
+		}
+	} else if env == Development {
+		// Even though code is same for both environments, we are keeping it separate to show that we can have different
+		// implementations for different environments. For example, we can have log levels set differently
+		logHandler := slog.NewJSONHandler(os.Stdout, nil).WithAttrs([]slog.Attr{slog.String("env", string(env))})
```
samlaf, line 25: why is it the same? is there a way to not show debug logs in production mode?

logging/zap_logger.go
```diff
@@ -61,26 +61,6 @@ func (z *ZapLogger) Fatal(msg string, tags ...any) {
...
-
-func (z *ZapLogger) Warnm(msg string, tags map[string]any) {
-	z.logger.Warn(msg, convertTagsToZapFields(tags)...)
-}
-
-func (z *ZapLogger) Errorm(msg string, tags map[string]any) {
-	z.logger.Error(msg, convertTagsToZapFields(tags)...)
-}
-
-func (z *ZapLogger) Fatalm(msg string, tags map[string]any) {
```
samlaf, line 80: you had previously removed these from the interface, but forgotten to delete them on the struct here, so cleaning up? Is this what's happening?

shrimalmadhur: yea! we don't use tags map anymore.

chainio/clients/avsregistry/writer.go
shrimalmadhur: no these are just formatting - I have a go lines plugin which makes sure lines don't go above 120.

logging/slog_logger.go
shrimalmadhur: yea - only `env` is different. since `slog` doesn't have different constructor. and basically dev and prod will both need structured log. ideally local will need more terminal like

samlaf: not following but it's fine with me since I won't be using slog for the time being, so someone else should request changes here if this is not fine for them, or we can change this later.

samlaf (approved):

logging/slog_logger.go
shrimalmadhur: oh wait. I read your comment wrong 🤦 . I think we can set the log level differently in prod, like only warn and above.

shrimalmadhur: actually by default it is `Info` only, to enable `debug` you have to explicitly set it to debug.

PR #114, opened 2024-02-13
we were printing tx hash of half-built transactions, which was pointless

samlaf:
Here's an example output from running avs-sync
image
Funnily enough, neither of the two txHash printed is the correct one, which was extremely confusing haha.

Changed it so that everywhere we only print txHash of txs that are successfully included in the chain.

### DIFF
--- a/chainio/clients/avsregistry/writer.go
+++ b/chainio/clients/avsregistry/writer.go
@@ -13,6 +13,7 @@ import (
 	"github.com/Layr-Labs/eigensdk-go/crypto/bls"
 	"github.com/Layr-Labs/eigensdk-go/logging"
 	"github.com/Layr-Labs/eigensdk-go/types"
+	sdkutils "github.com/Layr-Labs/eigensdk-go/utils"
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	gethcommon "github.com/ethereum/go-ethereum/common"
 	gethtypes "github.com/ethereum/go-ethereum/core/types"
@@ -255,8 +256,7 @@ func (w *AvsRegistryChainWriter) RegisterOperatorInQuorumWithAVSRegistryCoordina
 	if err != nil {
 		return nil, errors.New("failed to send tx with err: " + err.Error())
 	}
-	w.logger.Infof("tx hash: %s", tx.Hash().String())
-	w.logger.Info("registered operator with the AVS's registry coordinator")
+	w.logger.Info("succesfully registered operator with AVS registry coordinator", "tx hash", receipt.TxHash.String())
 	return receipt, nil
 }
 
@@ -265,7 +265,7 @@ func (w *AvsRegistryChainWriter) UpdateStakesOfEntireOperatorSetForQuorums(
 	operatorsPerQuorum [][]gethcommon.Address,
 	quorumNumbers []byte,
 ) (*gethtypes.Receipt, error) {
-	w.logger.Info("updating stakes for entire operator set", "quorumNumbers", quorumNumbers)
+	w.logger.Info("updating stakes for entire operator set", "quorumNumbers", sdkutils.ConvertQuorumsBytesToInts(quorumNumbers))
 	noSendTxOpts, err := w.txMgr.GetNoSendTxOpts()
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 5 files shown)


shrimalmadhur (approved):

chainio/clients/avsregistry/writer.go
```diff
@@ -254,7 +254,7 @@ func (w *AvsRegistryChainWriter) RegisterOperatorInQuorumWithAVSRegistryCoordina
 	if err != nil {
 		return nil, errors.New("failed to send tx with err: " + err.Error())
 	}
-	w.logger.Infof("tx hash: %s", tx.Hash().String())
+	w.logger.Info("tx succesfully included", "tx hash", receipt.TxHash.String())
 	w.logger.Info("registered operator with the AVS's registry coordinator")
```
uri-lightblocks, line 258: I would combine those two log records into a single record that includes all the data.

chainio/txmgr/txmgr.go
```diff
@@ -82,19 +82,21 @@ func NewSimpleTxManager(
 func (m *SimpleTxManager) Send(ctx context.Context, tx *types.Transaction) (*types.Receipt, error) {
 
 	// Estimate gas and nonce
-	m.log.Debug("Estimating gas and nonce", "tx", tx.Hash().Hex())
+	// can't print tx hash in logs because the tx changes below when we complete and sign it
+	// so the txHash is meaningless at this point
+	m.log.Debug("Estimating gas and nonce")
```
uri-lightblocks, line 87: Assuming there can be many calls to Send in parallel on different goroutines, how the debug log records without the transaction identifier will help in debugging?

shrimalmadhur: that's a good point. @samlaf should we keep the hash - and saying - it's an unsigned hash to make it clear?

samlaf: the problem is we'd want some kind of call tracing, but we if use the txhash inside and txreceipt.tx.hash outside, then the tracing "identifier" is not even unique..

chainio/clients/avsregistry/writer.go
samlaf: Done: [bdd6003](https://github.com/Layr-Labs/eigensdk-go/pull/114/commits/bdd60037541edc1deadf63d1ca1f2240dd4ade06)

chainio/txmgr/txmgr.go
samlaf: I would say let's cross that bridge when it's needed only. For now we won't be managing multiple wallets, so assuming a single tx is in flight at any point in time is safe I would think.

chainio/txmgr/txmgr.go
```diff
@@ -117,7 +120,7 @@ func (m *SimpleTxManager) Send(ctx context.Context, tx *types.Transaction) (*typ
 
 	tx, err = contract.RawTransact(opts, tx.Data())
 	if err != nil {
-		return nil, fmt.Errorf("send: failed to send txn: %w", err)
+		return nil, sdktypes.WrapError(fmt.Errorf("send: tx %v failed.", tx.Hash().String()), err)
```
samlaf, line 123: @shrimalmadhur added this to help with debugging. See https://github.com/Layr-Labs/avs-sync/issues/13

chainio/txmgr/txmgr.go
```diff
@@ -117,7 +120,7 @@ func (m *SimpleTxManager) Send(ctx context.Context, tx *types.Transaction) (*typ
...
 	}
 	if err != nil {
```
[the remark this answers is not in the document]
samlaf: yep thanks. [cf4ba6d](https://github.com/Layr-Labs/eigensdk-go/pull/114/commits/cf4ba6d4be392e79704899c83d6c91e03671e5b4)

chainio/txmgr/txmgr.go
shrimalmadhur: maybe eventually we wrap a transaction lifecyle with a unique identifier - that would be ideal to log and can work with multiple go routines, gas adjustments. it's okay for now

shrimalmadhur (approved):

shrimalmadhur (approved):

PR #115, opened 2024-02-13
added error logs for avs reader/writer clients

samlaf:
some avs teams who are forking our clients and remove some contracts such as the apkRegistry end up with bugs here, and the errors were not informative, so added logging lines for them.

### DIFF
--- a/chainio/clients/avsregistry/reader.go
+++ b/chainio/clients/avsregistry/reader.go
@@ -120,26 +120,26 @@ func BuildAvsRegistryChainReader(
 ) (*AvsRegistryChainReader, error) {
 	contractRegistryCoordinator, err := regcoord.NewContractRegistryCoordinator(registryCoordinatorAddr, ethClient)
 	if err != nil {
-		return nil, errors.Join(errors.New("Failed to create contractRegistryCoordinator"), err)
+		return nil, types.WrapError(errors.New("Failed to create contractRegistryCoordinator"), err)
 	}
 	blsApkRegistryAddr, err := contractRegistryCoordinator.BlsApkRegistry(&bind.CallOpts{})
 	if err != nil {
-		return nil, errors.Join(errors.New("Failed to get blsApkRegistryAddr"), err)
+		return nil, types.WrapError(errors.New("Failed to get blsApkRegistryAddr"), err)
 	}
 	stakeRegistryAddr, err := contractRegistryCoordinator.StakeRegistry(&bind.CallOpts{})
 	if err != nil {
-		return nil, errors.Join(errors.New("Failed to get stakeRegistryAddr"), err)
+		return nil, types.WrapError(errors.New("Failed to get stakeRegistryAddr"), err)
 	}
 	contractStakeRegistry, err := stakeregistry.NewContractStakeRegistry(stakeRegistryAddr, ethClient)
 	if err != nil {
-		return nil, errors.Join(errors.New("Failed to create contractStakeRegistry"), err)
+		return nil, types.WrapError(errors.New("Failed to create contractStakeRegistry"), err)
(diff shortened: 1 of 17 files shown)


chainio/clients/avsregistry/reader.go
```diff
@@ -119,25 +119,30 @@ func BuildAvsRegistryChainReader(
 ) (*AvsRegistryChainReader, error) {
 	contractRegistryCoordinator, err := regcoord.NewContractRegistryCoordinator(registryCoordinatorAddr, ethClient)
 	if err != nil {
+		logger.Error("Failed to create contractRegistryCoordinator", "err", err)
```
shrimalmadhur, line 122: should we rather wrap the error with informative text and then return it and not log here? reason it's client responsibility to log errors otherwise we have too many logs?

samlaf: Yeah I was thinking so too. But we've been doing it this way everywhere else so wanted to stay consistent. Shall we change it everywhere?

shrimalmadhur: yea we should maybe do this ? https://pkg.go.dev/errors#Join 

for ex: `errors.Join(errors.new("Failed to create contractRegistryCoordinator", err))`

uri-lightblocks: I agree with @shrimalmadhur , you should return the wrapped error and the client should handle it.

samlaf: > yea we should maybe do this ? https://pkg.go.dev/errors#Join
> 
> for ex: `errors.Join(errors.new("Failed to create contractRegistryCoordinator", err))`

join literally concatenates the error strings with a newline in between, such that printing them prints the 2 errors on two separate lines. Is that fine with you @uri-lightblocks ? I know you didn't like the newlines in avs-sync.

shrimalmadhur: I think new line is okay in the error object? not sure how will it behave in elk system when someone logs `err.Errror()` - I wonder if we should have custom error wrapping but I am scope creeping here 😅

samlaf: ok changed here: [856915e](https://github.com/Layr-Labs/eigensdk-go/pull/115/commits/856915ec9b42b9ecb4fd9666a73f9ad74c2a13fe)

samlaf: @uri-lightblocks tell me more man. How would you concatenate it? With a space in between? A tilda? A "-"? :)

samlaf: Let's get this merged. At least now we can easily change it whenever.

chainio/clients/avsregistry/reader.go
```diff
@@ -119,26 +120,26 @@ func BuildAvsRegistryChainReader(
 ) (*AvsRegistryChainReader, error) {
 	contractRegistryCoordinator, err := regcoord.NewContractRegistryCoordinator(registryCoordinatorAddr, ethClient)
 	if err != nil {
-		return nil, err
+		return nil, types.WrapError(errors.New("Failed to create contractRegistryCoordinator"), err)
```
shrimalmadhur, line 123: nit for @shrimalmadhur  for later 
- maybe this should not live in `types` should be in utils
- also interface should accept string and an error - so user doesn't need to do `errors.New`

not blocking this PR.

shrimalmadhur (approved):

chainio/clients/avsregistry/reader.go
shrimalmadhur: created https://github.com/Layr-Labs/eigensdk-go/issues/116

PR #117, opened 2024-02-16
Enable creating InstrumentedClient from existing

taco-paco:
…thods public

Fixes # .

### Motivation

Make some methods accessible for third-party use to avoid creating duplicate instances. Make new constructor method for Instrumented client for 3-party integration

### DIFF
--- a/chainio/clients/builder.go
+++ b/chainio/clients/builder.go
@@ -68,7 +68,7 @@ func BuildAll(
 
 	txMgr := txmgr.NewSimpleTxManager(ethHttpClient, logger, signerFn, signerAddr)
 	// creating EL clients: Reader, Writer and Subscriber
-	elChainReader, elChainWriter, err := config.buildElClients(
+	elChainReader, elChainWriter, err := config.BuildElClients(
 		ethHttpClient,
 		txMgr,
 		logger,
@@ -79,7 +79,7 @@ func BuildAll(
 	}
 
 	// creating AVS clients: Reader and Writer
-	avsRegistryChainReader, avsRegistryChainSubscriber, avsRegistryChainWriter, err := config.buildAvsClients(
+	avsRegistryChainReader, avsRegistryChainSubscriber, avsRegistryChainWriter, err := config.BuildAvsClients(
 		elChainReader,
 		ethHttpClient,
 		ethWsClient,
@@ -104,7 +104,7 @@ func BuildAll(
 
 }
 
-func (config *BuildAllConfig) buildElClients(
+func (config *BuildAllConfig) BuildElClients(
 	ethHttpClient eth.EthClient,
 	txMgr txmgr.TxManager,
 	logger logging.Logger,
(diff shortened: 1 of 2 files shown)


chainio/clients/eth/instrumented_client.go
```diff
@@ -42,6 +42,15 @@ func NewInstrumentedClient(rpcAddress string, rpcCallsCollector *rpccalls.Collec
 	}, nil
 }
 
+func NewInstrumentedClientFromClient(client *ethclient.Client, rpcCallsCollector *rpccalls.Collector) *InstrumentedClient {
```
samlaf, line 45: can you update `NewInstrumentedClient` to call this to prevent errors from duplicate code?

chainio/clients/builder.go
```diff
@@ -104,7 +104,7 @@ func BuildAll(
 
 }
 
-func (config *BuildAllConfig) buildElClients(
+func (config *BuildAllConfig) BuildElClients(
```
samlaf, line 107: Why do you need this method exposed? Personally find it a bit weird given that the BuildAllConfig takes in AVS contract addresses. I'd prefer we expose a BuildElClients that takes in a core eigenlayer contract address, such as the delegation manager address.

Also if you're already putting together the BuildAllConfig, might as well just build everything, even if you just need the EL clients.

samlaf (commented):
Thanks for this. Added some comments.
Curious in general what you're building and/or looking to do with these methods?
Are you part of an AVS team?

chainio/clients/builder.go
taco-paco: Its connected to `InstrumentedClient`. The reason to expose this is because `BuildAll` build clients using `eth.NewClient` which in our case just creates duplicate clients. I would like to be able to pass wrapped eth.Client(`InstrumentedClient`) to achieve results of `BuildAll` but with my own clients

taco-paco: > Its connected to `InstrumentedClient`. The reason to expose this is because `BuildAll` build clients using `eth.NewClient` which in our case just creates duplicate clients. I would like to be able to pass wrapped eth.Client(`InstrumentedClient`) to achieve results of `BuildAll` but with my own client instances

samlaf: I see, makes sense to expose these I guess for people to use them, although I'd still prefer we clear the TODO on top of Clients struct, and make a different BuildALL for people who want instrumented client and another for people who don't want metrics. Perhaps easiest is only to expose the building blocks like this, but my main concern was just that the config struct is then a bit confusing. I'm fine with merging these changes for now if it helps, but if you end up creating other useful high-level constructors, please create new PRs with those. :)

taco-paco: Sure) Will resolve conflicts in a few days.

taco-paco: Done

samlaf (approved):

PR #120, opened 2024-02-19
Improved key generation with factory pattern

DanielBoye:
Fixes a TODO in the code: 
https://github.com/Layr-Labs/eigensdk-go/blob/a5629b0ff188bbc876a1da705b5c4cd96ff8b7f0/cmd/egnkey/generate.go#L76

### Motivation
Updating to a factory pattern is smart to decouple object creation from its usage in our key generation logic. This change addresses the challenge of supporting multiple key types (ECDSA and BLS) in a scalable way. 

### Solution
The first I did was to include a KeyGenerator interface and two concrete implementations for ECDSA and BLS key types. The factory function, NewKeyGenerator, dynamically selects the appropriate generator based on input parameters, simplifying the generate function.

### Open questions
Would someone mind if we wrote unit tests for the newly implemented factory pattern in our key generation system, particularly focusing on the dynamic selection logic of the NewKeyGenerator function and ensuring each key generator (ECDSA and BLS) behaves as expected under various input conditions? Since as for now the `egnkey` does not have any test cases.

### DIFF
--- a/cmd/egnkey/generate.go
+++ b/cmd/egnkey/generate.go
@@ -63,92 +63,65 @@ It creates the following artifacts based on arguments
 	},
 }
 
-func generate(c *cli.Context) error {
-	keyType := c.String(KeyTypeFlag.Name)
-	if keyType != "ecdsa" && keyType != "bls" && keyType != "both" {
-		return cli.Exit("Invalid key type", 1)
-	}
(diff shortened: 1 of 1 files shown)


samlaf:
@shrimalmadhur can you take a look at this? I believe you wrote the egnkey util?

shrimalmadhur (approved):

PR #123, opened 2024-02-21
[HotWallet] Add Fireblocks client

ian-shim:
This PR adds an http client that makes requests to Fireblocks, supporting two methods for now: list contracts and contract call. 
The client will be used to implement a transaction sender that uses remote MPC.


chainio/clients/fireblocks/client.go
```diff
@@ -0,0 +1,162 @@
...
+type FireblocksClient interface {
+	ContractCall(ctx context.Context, body *ContractCallRequest) (*ContractCallResponse, error)
+	ListContracts(ctx context.Context) ([]WhitelistedContract, error)
+}
+
+type fireblocksClient struct {
+	apiKey     string
+	privateKey *rsa.PrivateKey
+	baseURL    string
+	timeout    time.Duration
```
samlaf, line 33: timeout is used in both `http.Client{Timeout: timeout}` as well as the jwt token expiration. Feels like those two things are completely unrelated and should be different parameters? Unless I'm missing something..

chainio/clients/fireblocks/client.go
```diff
@@ -0,0 +1,162 @@
...
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/golang-jwt/jwt"
+	"github.com/google/uuid"
+)
+
+type FireblocksTxID string
+
+type FireblocksClient interface {
+	ContractCall(ctx context.Context, body *ContractCallRequest) (*ContractCallResponse, error)
+	ListContracts(ctx context.Context) ([]WhitelistedContract, error)
```
samlaf, line 26: Could you detail here how ListContracts is supposed to be used. Unclear what its goal is and why we need it.
Since there's no parameters, where are the WhitelistedContracts coming from? Is this for OFAC compliance purposes... or?

chainio/clients/fireblocks/client.go
```diff
@@ -0,0 +1,162 @@
...
+	"github.com/Layr-Labs/eigensdk-go/logging"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/golang-jwt/jwt"
+	"github.com/google/uuid"
+)
+
+type FireblocksTxID string
+
+type FireblocksClient interface {
+	ContractCall(ctx context.Context, body *ContractCallRequest) (*ContractCallResponse, error)
```
samlaf, line 25: Is this really a contractCall (just reading data)? Or is it a send transaction?

I'm guessing its actually signing and sending a tx as that's the only goal of fireblocks right? I see that fireblocks itself has named this operator CONTRACT_CALL in their api, any idea why? It's confusing.

chainio/clients/mocks/fireblocks.go
samlaf, line 1: We've been using https://github.com/uber-go/mock in the sdk. Is there any reason why you absolutely need testify/mock? If not, I'd prefer we only use one framework. uber mock auto generates mock clients just from the interface, so just need to add a `generate` command in the gen.go file to generate it, instead of having to write it by hand.

samlaf (commented):
Added first set of comments. I'm still unsure how this fits with the remote signers framework @shrimalmadhur had created. Is it incompatible, or do you plan on also having this fireblocks client implement the remote signer interface? I'll defer to @shrimalmadhur here as he's the expert who designed and been thinking about remote signing.

chainio/clients/fireblocks/client.go
```diff
@@ -0,0 +1,162 @@
...
+	ID     string  `json:"id"`
+	Name   string  `json:"name"`
+	Assets []Asset `json:"assets"`
+}
+
+func NewFireblocksClient(apiKey string, secretKey []byte, baseURL string, timeout time.Duration, logger logging.Logger) (FireblocksClient, error) {
+	c := http.Client{Timeout: timeout}
+	privateKey, err := jwt.ParseRSAPrivateKeyFromPEM(secretKey)
+	if err != nil {
+		return nil, fmt.Errorf("error parsing RSA private key: %w", err)
```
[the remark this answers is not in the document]
samlaf: Did a quick search. Seems like `errors.Join()` is the new way since go 1.20 to combine errors (see for eg https://lukas.zapletalovi.com/posts/2022/wrapping-multiple-errors/). Wrapping would wrap errors (one contains the other), whereas join creates an error with a list of error strings. Still not 100% sure what the difference is, but it seems like join is easier to deal with and implement `is` to check if a combined error contains some error.

chainio/clients/fireblocks/client.go
samlaf: I see. Can't you request a JWT token for an hour or so and then use it for multiple calls? Or are you forced to request a new one on every call? Seems very inneficient.

chainio/clients/clients_test.go
```diff
@@ -0,0 +1,11 @@
+package clients_test
+
+import (
+	"os"
+	"testing"
+)
+
+func TestMain(m *testing.M) {
```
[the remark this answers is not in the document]
samlaf: We've been running all tests via `go test ./...`. Each test can have its own setup/teardown routines. Is this just a global setup/teardown? Are tests still run via `go test ./...`?

chainio/clients/fireblocks/client.go
samlaf: Makes sense. Can you just add this as a comment above the Method?

chainio/clients/fireblocks/client.go
samlaf: This is so weird. See for eg https://ethereum.stackexchange.com/questions/765/what-is-the-difference-between-a-transaction-and-a-call
Can we at least document this as a comment above the method?

chainio/clients/mocks/fireblocks.go
samlaf: Please 🙏

chainio/clients/clients_test.go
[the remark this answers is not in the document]
ian-shim: Yes, it can be run the same way. It just allows setup/teardown for tests at the package level.

chainio/clients/fireblocks/client.go
ian-shim: Yeah, we had to ask a few clarifying questions to learn the behavior for this method. I'll add a comment

chainio/clients/fireblocks/client.go
ian-shim: I think you pretty much have to generate a new token for every call since the token signs the path and body.

chainio/clients/fireblocks/client.go
```diff
@@ -0,0 +1,168 @@
...
+
+type FireblocksTxID string
+
+type FireblocksClient interface {
+	// ContractCall makes a ContractCall request to the Fireblocks API.
+	// It signs and broadcasts a transaction and returns the transaction ID and status.
+	// ref: https://developers.fireblocks.com/reference/post_transactions
+	ContractCall(ctx context.Context, body *ContractCallRequest) (*ContractCallResponse, error)
+	// ListContracts makes a ListContracts request to the Fireblocks API
+	// It returns a list of whitelisted contracts and their assets for the account. This call is necessary to get the contract ID for a whitelisted contract.
```
samlaf, line 30: is the contract ID the `sourceaccountId` that is passed to `NewContractCallRequest`? If so can you just change the sentence to
```
This call is used to get the contract ID for a whitelisted contract, which is needed by NewContractCallRequest before making a ContractCall
```
or somethng like this?

chainio/clients/mocks/fireblocks.go
samlaf, line 1: Did you generate this manually?
Can you instead add a line for fireblocks at the bottom of https://github.com/Layr-Labs/eigensdk-go/blob/master/chainio/gen.go? This is where we put all the mock generation commands, so we can generate/update all of them at once with `make mocks`

samlaf (commented):
last minor nits

chainio/clients/mocks/fireblocks.go
ian-shim: Yeah I manually generated it. I'll add it in that file!

samlaf (approved):
🚀🚀🚀

PR #124, opened 2024-02-22
[HotWallet] Add `txsender`

ian-shim:
This PR introduces a new interface `txsender`, which supports a few methods:
- `SendTransaction`: signs a raw transaction and broadcasts it to the network
- `GetTransactionReceipt `: retrieves the receipt if available
- ~~`SenderAddress`: address of the transaction sender~~

It also refactors the `TxManager` to use `txsender` instead of signing/sending the transaction itself. 

This interface will be used to implement remote MPC txsender that uses Fireblocks API.


chainio/txmgr/txmgr.go
```diff
@@ -142,93 +101,3 @@ func (m *SimpleTxManager) GetNoSendTxOpts() (*bind.TransactOpts, error) {
 	}
 	return noSendTxOpts, nil
```
samlaf, line 102: is this function `GetNoSendTxOpts` still being used?

chainio/clients/txsender/privateKeyTxSender.go
```diff
@@ -0,0 +1,180 @@
...
+	if err != nil {
+		return "", sdktypes.WrapError(fmt.Errorf("send: tx %v failed.", tx.Hash().String()), err)
+	}
+
+	return "", nil
+}
+
+func (t *privateKeyTxSender) WaitForTransactionReceipt(ctx context.Context, txID TxID) (*types.Receipt, error) {
+	txHash := common.HexToHash(txID)
+	queryTicker := time.NewTicker(2 * time.Second)
```
samlaf, line 103: some AVS team mentioned that we should be sleeping more smartly, since they just get a bunch of "waiting for receipt" logs that clutter the output. Since blocks are 12s on mainnet and 15s on goerli, we could potentially add some logic to sleep until the next block before retrying etc.

I'm totally fine with not including this logic right now, but can we just add a TODO comment to remember to do this eventually (or some external contributor might do it for us :) )

chainio/clients/txsender/privateKeyTxSender.go
```diff
@@ -0,0 +1,180 @@
...
+		// cache the contract for later use
+		t.contracts[*tx.To()] = contract
+	}
+
+	tx, err = contract.RawTransact(opts, tx.Data())
+	if err != nil {
+		return "", sdktypes.WrapError(fmt.Errorf("send: tx %v failed.", tx.Hash().String()), err)
+	}
+
+	return "", nil
```
samlaf, line 98: why are you returning "" here? Won't this not work with the waitForTransactionReceipt logic?

chainio/clients/txsender/privateKeyTxSender.go
```diff
@@ -0,0 +1,180 @@
...
+		case <-queryTicker.C:
+			if receipt := t.queryReceipt(ctx, txHash); receipt != nil {
+				return receipt, nil
+			}
+		}
+	}
+}
+
+func (t *privateKeyTxSender) queryReceipt(ctx context.Context, txHash common.Hash) *types.Receipt {
+	receipt, err := t.ethClient.TransactionReceipt(ctx, txHash)
```
samlaf, line 118: what happens here when txHash is ""?

chainio/clients/txsender/privateKeyTxSender.go
samlaf: guessing it should be `tx.Hash()` or something?

samlaf (commented):
So we're left with these 2 interfaces
```
type TxSender interface {
	SendTransaction(ctx context.Context, tx *types.Transaction) (TxID, error)
	WaitForTransactionReceipt(ctx context.Context, txID TxID) (*types.Receipt, error)
	SenderAddress() common.Address
}

type TxManager interface {
	Send(ctx context.Context, tx *types.Transaction) (*types.Receipt, error)
	GetNoSendTxOpts() (*bind.TransactOpts, error)
}
```

which look very very similar to me. Do we still need the txmgr? Unclear to me that they don't serve the same purpose?

chainio/txmgr/txmgr.go
ian-shim: Yeah, in avsregistry writer and elcontracts writer

chainio/clients/txsender/privateKeyTxSender.go
ian-shim: ah yes! thanks for catching this

chainio/clients/txsender/privateKeyTxSender.go
ian-shim: empty `txHash` is `0x0000000000...`, so it would try to get a receipt for transaction `0x0000000000...` and fail

ian-shim (commented):
> which look very very similar to me. Do we still need the txmgr? Unclear to me that they don't serve the same purpose?

I agree that they look very similar from the interface/naming. Maybe we can clarify the naming of the two interfaces, but my understanding is the two have different purposes. `txmgr` "manages" transactions in that it would manage the nonces, regas logic in case tx is stalled, etc. while `txsender` supports more primitive functionalities like sending a raw transaction. So `txmgr` would use `txsender` to achieve the high level functionalities

chainio/clients/txsender/privateKeyTxSender.go
samlaf: makes sense, was asking in the context of the other bug, but I see you fixed that.

chainio/txmgr/txmgr.go
samlaf: Ok thanks for the sync yesterday, was helpful in understanding this. The part I don't understand now is how txsender can be used now with a privatekey signer. The flow in the writer clients is something like:
```
// get the noSendTxOpts that is passed to abigen functions
noSendTxOpts, err := w.txMgr.GetNoSendTxOpts()
// pass the noSendTxOpts to abigen to construct the tx
tx, err := w.someAbiGenContract.someAbiGenFunction(noSendTxOpts, args...)
// send the tx
receipt, err := w.txMgr.Send(ctx, tx)
```

Given that a txsender still takes a tx of type `*types.Transaction`, it seems like this would still need to be constructed via abigen, which requires calling `txMgr.GetNoSendTxOpts()` the same way? So it looks to me like this txsender interface is not self contained and still requires using a txMgr.

ian-shim: The existing writers will still be using txMgr to interact with the chain since txsender isn't meant to replace txMgr. It just provides more primitive methods that txMgr uses to interact with the chain

samlaf: So txsender should not be used by itself then? Ever?

ian-shim: Only when you need the basic broadcasting of a transaction or retrieving a transaction status.
If you want transaction monitoring/gas updates, txmgr should handle it.

chainio/txmgr/txmgr.go
```diff
@@ -48,29 +42,30 @@ type EthBackend interface {
...
+	sender   common.Address
 }
 
 var _ TxManager = (*SimpleTxManager)(nil)
 
 // NewSimpleTxManager creates a new simpleTxManager which can be used
 // to send a transaction to smart contracts on the Ethereum node
 func NewSimpleTxManager(
+	txSender txsender.TxSender,
 	backend EthBackend,
```
samlaf, line 58: Think we should get rid of this ethbackend interface now, and just use eth.Client interface everywhere (pretty sure eth.Client contains all of these EthBackend methods).

Can you just add a TODO here above `backend EthBackend` to investigate whether we can just replace it by the eth.Client. Let's do it in another PR though to not make this PR too big.

chainio/clients/txsender/privatekey_txsender.go
```diff
@@ -0,0 +1,149 @@
...
+	ethClient eth.Client
+	address   common.Address
+	signerFn  signerv2.SignerFn
+	logger    logging.Logger
+
+	// cache
+	contracts map[common.Address]*bind.BoundContract
+}
+
+func NewPrivateKeyTxSender(ethClient eth.Client, chainID *big.Int, ecdsaPrivateKey *ecdsa.PrivateKey, logger logging.Logger) (TxSender, error) {
```
shrimalmadhur, line 36: SignerV2 should be input to this function. not ecdsa private key. since if we have web3 signer we won't have access to it. we should create signer before calling it based on SignerConfig and pass here. or you are thinking we would define `NewKeyStoreTxSender` and `NewWeb3TxSender`?

shrimalmadhur: these should be Signers which we have and txSender constructor should just accept signerV2 itself

chainio/clients/txsender/txsender.go
```diff
@@ -0,0 +1,15 @@
...
+
+import (
+	"context"
+
+	"github.com/ethereum/go-ethereum/core/types"
+)
+
+type TxID = string
+
+// TxSender is an interface for signing and sending transactions to the network
```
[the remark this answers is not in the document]
ian-shim: To be clear, this interface can be used to build a sender that sends tx to a private mempool, but that's not the primary use case. At least for hot wallet implementation, this interface is necessary because transactions are broadcasted via external API, not eth client. I'll add a comment documenting this

chainio/clients/txsender/privatekey_txsender.go
ian-shim: I was thinking we'd have separate txsender implementations for each signer and that's why I called it `PrivateKeyTxSender` because it uses the `PrivateKeySignerFn`. But I don't see why we need separate txsender per signer method, so I'll consolidate it and make it take SignerV2 as an input.

shrimalmadhur: yep. I think interface should be same since it's same behavior of transaction sending which is the TxSender/Wallet's job. It should not have knowledge of signer for this.

samlaf (approved):
LGTM!

PR #127, opened 2024-02-24
Updates to slog to make it the default feature-full logger

samlaf:
Started looking into slog:
- https://go.dev/blog/slog
- https://www.youtube.com/watch?v=8rnI2xLrdeM

and agree with @dmanc at this point that we should probably start adopting slog as our standard logger everywhere. Been playing with it a bit and our current interface is super limiting wrt what can be done with slog (for eg can't pass a context to logging functions right now, which can be used to also log traceIds). The blog above mentions this
image

This draft PR is a wip for adding features to the slogger itself. But we'll eventually need to find a way to adapt our main interface, probably in a next PR. This will require more discussions.

Features added thus far:
- Only print required log level (production logger will only print info and above logs, whereas the development one will also print debug logs) ~~[4c03592](https://github.com/Layr-Labs/eigensdk-go/pull/127/commits/4c0359225bdb90f6a83c1ae0116f190dd7b42992)~~ Found a much easier way to do this: [df441e7](https://github.com/Layr-Labs/eigensdk-go/pull/127/commits/df441e7ad9bb8b835ef3295e8b86c5be4cfbb7c6)
- Separate constructor into two, one for TextHandler and one for JsonHandler (I'm personally becoming a big fan of golang's key-value text format) [52c50b7](https://github.com/Layr-Labs/eigensdk-go/pull/127/commits/52c50b741847ac6fc2478d61b5ca280c84e9b7f3)
- Print proper source file information (see image below*) [52c50b7](https://github.com/Layr-Labs/eigensdk-go/pull/127/commits/52c50b741847ac6fc2478d61b5ca280c84e9b7f3)
- make the slog logger an embedded struct (instead of named field) so that we have access to its method. Also return the SLogger struct instead of Logger interface from constructors so that users can use those methods (.WithGroup for eg) [d5b616a](https://github.com/Layr-Labs/eigensdk-go/pull/127/commits/d5b616ab162bf0415e9a9b3363f2c0a7761eeb63)

* because of our log interface and structs that wrap the underlying logger, the srouce information getting printed was always the wrapper location, instead of its caller, which is useless:
image

### DIFF
--- a/logging/slog_logger.go
+++ b/logging/slog_logger.go
@@ -1,6 +1,7 @@
 package logging
 
 import (
+	"context"
 	"fmt"
 	"log/slog"
 	"os"
@@ -14,16 +15,16 @@ var _ Logger = (*SLogger)(nil)
 
 func NewSlogLogger(env LogLevel) Logger {
 	if env == Production {
-		logHandler := slog.NewJSONHandler(os.Stdout, nil).WithAttrs([]slog.Attr{slog.String("env", string(env))})
-		logger := slog.New(logHandler)
+		jsonHandler := slog.NewJSONHandler(os.Stdout, nil).WithAttrs([]slog.Attr{slog.String("env", string(env))})
+		levelHandler := NewLevelHandler(slog.LevelInfo, jsonHandler) // production logs only info and above
+		logger := slog.New(levelHandler)
 		return &SLogger{
 			logger: logger,
 		}
 	} else if env == Development {
-		// Even though code is same for both environments, we are keeping it separate to show that we can have different
-		// implementations for different environments. For example, we can have log levels set differently
-		logHandler := slog.NewJSONHandler(os.Stdout, nil).WithAttrs([]slog.Attr{slog.String("env", string(env))})
-		logger := slog.New(logHandler)
+		jsonHandler := slog.NewJSONHandler(os.Stdout, nil).WithAttrs([]slog.Attr{slog.String("env", string(env))})
+		levelHandler := NewLevelHandler(slog.LevelDebug, jsonHandler) // development logs everything
+		logger := slog.New(levelHandler)
 		return &SLogger{
 			logger: logger,
 		}
@@ -72,4 +73,49 @@ func (s SLogger) Errorf(template string, args ...interface{}) {
 func (s SLogger) Fatalf(template string, args ...interface{}) {
 	s.logger.Error(fmt.Sprintf(template, args...))
 	os.Exit(1)
+}
+
+// Taken from https://pkg.go.dev/log/slog@master#example-Handler-LevelHandler
+// A LevelHandler wraps a Handler with an Enabled method
+// that returns false for levels below a minimum.
+// This permits our logger to only log messages of a certain level or higher
+type LevelHandler struct {
+	level   slog.Leveler
+	handler slog.Handler
+}
+
+// NewLevelHandler returns a LevelHandler with the given level.
+// All methods except Enabled delegate to h.
+func NewLevelHandler(level slog.Leveler, h slog.Handler) *LevelHandler {
+	// Optimization: avoid chains of LevelHandlers.
+	if lh, ok := h.(*LevelHandler); ok {
+		h = lh.Handler()
+	}
+	return &LevelHandler{level, h}
+}
+
+// Enabled implements Handler.Enabled by reporting whether
+// level is at least as large as h's level.
+func (h *LevelHandler) Enabled(_ context.Context, level slog.Level) bool {
+	return level >= h.level.Level()
+}
+
+// Handle implements Handler.Handle.
+func (h *LevelHandler) Handle(ctx context.Context, r slog.Record) error {
+	return h.handler.Handle(ctx, r)
+}
+
+// WithAttrs implements Handler.WithAttrs.
+func (h *LevelHandler) WithAttrs(attrs []slog.Attr) slog.Handler {
+	return NewLevelHandler(h.level, h.handler.WithAttrs(attrs))
+}
+
+// WithGroup implements Handler.WithGroup.
+func (h *LevelHandler) WithGroup(name string) slog.Handler {
+	return NewLevelHandler(h.level, h.handler.WithGroup(name))
+}
+
+// Handler returns the Handler wrapped by h.
+func (h *LevelHandler) Handler() slog.Handler {
+	return h.handler
 }
--- a/logging/slog_logger.go
+++ b/logging/slog_logger.go
@@ -3,6 +3,7 @@ package logging
 import (
 	"context"
 	"fmt"
+	"io"
 	"log/slog"
 	"os"
 	"runtime"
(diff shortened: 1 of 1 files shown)


logging/slog_logger.go
```diff
@@ -1,75 +1,128 @@
...
-	logger *slog.Logger
+	*slog.Logger
 }
 
 var _ Logger = (*SLogger)(nil)
 
-func NewSlogLogger(env LogLevel) Logger {
+func NewSlogTextLogger(env LogLevel) *SLogger {
+	var handler slog.Handler
+	if env == Production {
```
dmanc, line 27: Not sure if it makes sense to filter out logs based on production vs development. Sometimes we need to output debug logs in our production environment so we can properly debug production issues.

dmanc: Wouldn't it make sense to just make env the levels themselves from slog?

dmanc (commented):
Another thing that could be nice is to have the option to write logs to a file in addition to stdout.

logging/slog_logger.go
samlaf: Yeah we should.. we were using the log level names from zapper, no idea why they only have prod and dev. I would vote to just switch everything to slog, but this might break some client code. You fine with this @shrimalmadhur ?
Honestly eventually we should probably just migrate our zap logger to using the slog handler https://github.com/uber-go/zap/tree/master/exp/zapslog so that everything's consistent. Thoughts @shrimalmadhur ?

samlaf: @dmanc are you currently switching the log level dynamically? That's an option we could add, by exposing some REST endpoint or whatever.

samlaf: @dmanc changed the constructors in [bded386](https://github.com/Layr-Labs/eigensdk-go/pull/127/commits/bded3869144bcfd7124ac19ebd63ac753d9f0b19)
Let me know what you think

samlaf:
> Another thing that could be nice is to have the option to write logs to a file in addition to stdout.

Check [bded386](https://github.com/Layr-Labs/eigensdk-go/pull/127/commits/bded3869144bcfd7124ac19ebd63ac753d9f0b19)
Made the `io.Writer` an argument to the constructors. So to write to both stdout and a file, you can just pass a `io.MultiWriter(os.Stdout, file)`.

logging/slog_logger.go
dmanc: We're not switching dynamically during runtime but that sounds like a good idea! Right now it's just config and we need to restart.

samlaf: How helpful would this dynamic switching be? Like how important is it? Aka can we merge this without?

dmanc: Not important right now

logging/slog_logger.go
```diff
@@ -1,75 +1,119 @@
...
-		}
-	} else {
-		panic(fmt.Sprintf("Unknown environment. Expected %s or %s. Received %s.", Development, Production, env))
+// NewSlogTextLogger creates a new SLogger with a text handler
+//
+//	outputWriter is the writer to write the logs to (typically os.Stdout,
+//	  but can also use a io.MultiWriter(os.Stdout, file) to write to multiple outputs)
+//	logLevel is the minimum level to log
+//	logSource if true, adds source information to the log
+func NewSlogTextLogger(outputWriter io.Writer, logLevel slog.Level, logSource bool) *SLogger {
```
shrimalmadhur, line 32: I feel this should expect a log config which should have: env, log level etc - whatever is needed.

logging/slog_logger.go
```diff
@@ -1,75 +1,119 @@
...
 	"os"
+	"runtime"
+	"time"
+)
+
+type SlogHandlerType string
+
+const (
+	SLogHandlerTypeJson     SlogHandlerType = "json" // prints debug and above
+	SLogHandlerTypeJsonText SlogHandlerType = "text" // prints info and above
```
ian-shim, line 17: Are they used?

logging/slog_logger.go
```diff
@@ -1,75 +1,119 @@
...
 
 func (s SLogger) Fatal(msg string, tags ...any) {
-	s.logger.Error(msg, tags...)
+	s.logCorrectSource(slog.LevelError, msg, tags...)
 	os.Exit(1)
 }
 
+// logCorrectSource logs the message with the correct source information
+// adapted from https://pkg.go.dev/log/slog#example-package-Wrapping
+func (s SLogger) logCorrectSource(level slog.Level, msg string, tags ...any) {
```
ian-shim, line 83: do we always want this feature?
should this be an option the user configures?

ian-shim (approved):

logging/slog_logger.go
shrimalmadhur: I think it is an user option - https://github.com/Layr-Labs/eigensdk-go/pull/127/files#diff-54eaa5e5eaaf4b31c21b1f44503ed097a88564b92a04a4f35a17ce5240612b6cR52 ?

logging/slog_logger.go
samlaf: Ok will resolve this thread then. Let's revisit this dynamic switching later.

logging/slog_logger.go
samlaf: Indeed. resolving

logging/slog_logger.go
samlaf: Nope, good catch. Originally wanted this as an option to the constructor, but opted for 2 separate constructors. Removed in [c276964](https://github.com/Layr-Labs/eigensdk-go/pull/127/commits/c2769642573052270294ccbdc108357f311b3118)

logging/slog_logger.go
samlaf: Shall we take the entire `slog.HandlerOptions` as input then?
So `NewSlogTextLogger(outputWriter io.Writer, handerOptions slog.HandlerOptions)`

logging/slog_logger.go
```diff
@@ -1,75 +1,112 @@
...
-		panic(fmt.Sprintf("Unknown environment. Expected %s or %s. Received %s.", Development, Production, env))
+// NewSlogTextLogger creates a new SLogger with a text handler
+//
+//	outputWriter is the writer to write the logs to (typically os.Stdout,
+//	  but can also use a io.MultiWriter(os.Stdout, file) to write to multiple outputs)
+//	logLevel is the minimum level to log
+//	logSource if true, adds source information to the log
+func NewSlogTextLogger(outputWriter io.Writer, logLevel slog.Level, logSource bool) *SLogger {
+	handler := slog.NewTextHandler(outputWriter, &slog.HandlerOptions{
+		Level:     logLevel, // prints debug and above
```
ian-shim, line 27: nit: `// prints this level and above`

logging/slog_logger.go
ian-shim: I think that makes sense or something like
```
type SLoggerConfig struct {
 slog.HandlerOptions
 OutputWriter io.Writer
}
```

@shrimalmadhur wdyt?

logging/slog_logger.go
samlaf: [78e783b](https://github.com/Layr-Labs/eigensdk-go/pull/127/commits/78e783be8562d48f38ca5483ea7f3e5fd5d1a53b)

logging/slog_logger.go
samlaf: See [2bfeb3f](https://github.com/Layr-Labs/eigensdk-go/pull/127/commits/2bfeb3f7f7f662167474e29c4f7a243ffe0aa3fa)

ian-shim (approved):

PR #131, opened 2024-02-26
[HotWallet] Implement Fireblocks TxSender

ian-shim:
This PR introduces Fireblocks TxSender which uses fireblocks client to implement an MPC tx sender

### DIFF
--- a/chainio/clients/eth/client.go
+++ b/chainio/clients/eth/client.go
@@ -3,7 +3,6 @@ package eth
 import (
 	"context"
 	"math/big"
-	"time"
 
 	"github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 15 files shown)


chainio/clients/eth/client.go
```diff
@@ -11,7 +11,7 @@ import (
 	"github.com/ethereum/go-ethereum/ethclient"
 )
 
-type gethClient interface {
+type GethClient interface {
```
samlaf, line 14: This made me realize, with this new txsender and txmanager etc, is the WaitForTransactionReceipt function below on our EthClient interface really the right place? Or should we expose it on the txsender or txmgr etc? cc @shrimalmadhur

samlaf: Oh yeah txsender exposes the WaitForTransactionReceipt. So let's get rid of the EthClient interface below then?

I'd also vote to name this interface EthClient instead of GethClient, because this works with any EL client, not only geth.

ian-shim: done

samlaf: Lgtm but I want to make sure @shrimalmadhur sees this interface breaking change and approves before merging.

PR #134, opened 2024-02-29
[HotWallet] Handle Transaction Replacement at Fireblock Wallet

ian-shim:
To replace a transaction with Fireblocks, you need to provide an explicit transaction hash to replace (`replaceTxByHash`), and use this parameter to send a new transaction (`ContractCall`). Ideally, to be compatible with eth client interface, this parameter is automatically set by looking at the nonce, checking if the nonce has been used, and adding the `replaceTxByHash` param if necessary.
This PR makes fireblocks wallet keep track of nonces and txID to look up the transaction hashes of the respective nonces, then setting the `replaceTxByHash` param if necessary.


chainio/clients/wallet/fireblocks_wallet.go
```diff
@@ -18,12 +19,20 @@ import (
 var _ Wallet = (*fireblocksWallet)(nil)
 
 type fireblocksWallet struct {
+	mu sync.Mutex
```
samlaf, line 22: what is this mutex for? to protect concurrent writes to nonceToTxID and txIDToNonce? Can we name the mutex more explicitly?
Also write a small explanation as to why its needed. Is the idea that I could be trying to send multiple txs from different threads? Unclear to me why this would happen.

samlaf: oh just realized there's also concurrent access between send and gettxreceipt

chainio/clients/wallet/fireblocks_wallet.go
```diff
@@ -118,18 +130,40 @@ func (t *fireblocksWallet) SendTransaction(ctx context.Context, tx *types.Transa
 	if err != nil {
 		return "", fmt.Errorf("error getting whitelisted contract %s: %w", tx.To().Hex(), err)
 	}
+
+	t.mu.Lock()
+	defer t.mu.Unlock()
+	// if the nonce is already in the map, it means that the transaction was already submitted
+	// we need to get the replacement transaction hash and use it as the replaceTxByHash parameter
+	replaceTxByHash := ""
+	nonce := tx.Nonce()
```
samlaf, line 139: so the idea is that I would send the same tx but with a gas price increase? and the nonce here would be the same as the previous tx because it still hasn't been mined so when abigen creates the tx it gets the old nonce?

samlaf (approved):
mostly LGTM. Added some questions which are mostly for my understanding.

chainio/clients/fireblocks/contract_call.go
```diff
@@ -35,6 +35,9 @@ type ContractCallRequest struct {
 	Destination     account              `json:"destination"`
 	Amount          string               `json:"amount"`
 	ExtraParameters extraParams          `json:"extraParameters"`
+	// In case a transaction is stuck, specify the hash of the stuck transaction to replace it
+	// by this transaction with a higher fee, or to replace it with this transaction with a zero fee and drop it from the blockchain.
```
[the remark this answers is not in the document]
ian-shim: Ah yeah normally the transaction should be zero value transfer with the same nonce to cancel the previous one. I copy pasted this comment from their doc, but I'm not sure if they meant actual zero fee or if it's a typo. I'll ask them

Screenshot 2024-03-01 at 6 12 38 PM

chainio/clients/wallet/fireblocks_wallet.go
ian-shim: yup I'll add comment

chainio/clients/wallet/fireblocks_wallet.go
ian-shim: Yes, this method doesn't change any parameters, so the increased gas price should be specified in the tx. This is the same interface you would normally use chain client to replace a transaction.

samlaf (approved):
LGTM, other than that one unclear comment that's still open. Feel free to merge though if this unblocks you.

ian-shim:
> LGTM, other than that one unclear comment that's still open. Feel free to merge though if this unblocks you.

Yes, I'll merge this and integrate with existing wallets to see if the replacement logic works as intended. 
Not sure what got force pushed, but I think it needs an approval again

samlaf (approved):

PR #142, opened 2024-03-07
Use default folder for storing keys

DanielBoye:
Fixes a TODO:
https://github.com/Layr-Labs/eigensdk-go/blob/bfab929c131da84358295100cdac39dc5cbfe1a7/crypto/utils/batch_key.go#L11

### Motivation
Addresses the first TODO in the `crypto/utils/batch_key.go` file, which involves moving constants to a common constants file. The specific constant being addressed is `DefaultKeyFolder`, which is the default folder for storing keys. The motivation behind this change is to promote code reusability and maintainability by consolidating common constants in a shared location.

### Solution
The solution involves relocating the `DefaultKeyFolder` constant from the `batch_key.go` file to a common constants file. This common file, which is yet to be created, will serve as a central repository for shared constants accessible by both the `utils` and `cmd` packages. No changes in functionality are introduced, and the code remains consistent with the original design.

### Open questions

### DIFF
--- a/crypto/utils/batch_key.go
+++ b/crypto/utils/batch_key.go
@@ -8,8 +8,7 @@ import (
 )
 
 const (
-	// TODO: move to a common constants file which
-	// can be used by utils and cmd
+	// DefaultKeyFolder is the default folder for storing keys.
 	DefaultKeyFolder  = "keys"
 	PasswordFile      = "password.txt"
 	PrivateKeyHexFile = "private_key_hex.txt"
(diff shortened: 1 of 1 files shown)


crypto/utils/batch_key.go
```diff
@@ -8,8 +8,7 @@ import (
 )
 
 const (
-	// TODO: move to a common constants file which
-	// can be used by utils and cmd
```
samlaf, line 12: Can you also change `/cmd/egnkey/generate.go` to use these constants from utils instead of redefining it themselves. I think this is what this TODO was meant to say. I tried just now and it seems there's no circular dependency so not sure why we redefined. Change should be straightforward.

samlaf (changes requested):

samlaf:
@DanielBoye really sorry about this we dropped the ball on your PRs. If you want to reopen I'll review them.

PR #146, opened 2024-03-09
Add basic types with slog logvalues

samlaf:
Some of our core types were hard to read because they were logged as base64 encoded values, or just long byte strings.
This PR creates a few new basic types that have slog LogValues to print them in a more human-readable way.

### DIFF
--- a/types/operator.go
+++ b/types/operator.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"log/slog"
 	"math/big"
 	"net/http"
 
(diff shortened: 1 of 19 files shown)


chainio/clients/builder.go
```diff
@@ -68,7 +69,13 @@ func BuildAll(
 		return nil, types.WrapError(errors.New("Failed to create Eth WS client"), err)
 	}
 
-	signerV2, addr, err := signerv2.SignerFromConfig(signerv2.Config{PrivateKey: ecdsaPrivateKey}, big.NewInt(1))
+	rpcCtx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
shrimalmadhur, line 72: you need to rebase?

types/operator.go
```diff
@@ -88,9 +89,50 @@ type StakeAmount = *big.Int
...
-type QuorumNum = uint8
-type QuorumThresholdPercentage = uint32
+type OperatorId = Bytes32
+type QuorumNums []QuorumNum
+
+func (q QuorumNums) LogValue() slog.Value {
+	return slog.StringValue(fmt.Sprintf("%v", q))
+}
+
+func (q QuorumNums) Underlying() []uint8 {
```
shrimalmadhur, line 99: can there be a better name? when I read `Underlying()` in code, I am like what does it even mean? like the above `GetUnderlyingBytes` is self explanatory.

chainio/clients/builder.go
samlaf: I don't think so.. why are you saying this?

types/operator.go
samlaf: It's not bytes though, it returns the underlying type.
The idea is that eigensdk wraps normal types like `type QuorumNum int8` so that we can define methods on QuorumNum. But the contract bindings expect the unwrapped (underlying) type, hence the name `underlying()`. I think it makes sense?
Could also potentially call it `Unwrap()`?

shrimalmadhur: ye `Unwrap` maybe is better. but i just feel `underlying` doesn't say anything. `underlyingType`?

chainio/clients/builder.go
shrimalmadhur: you pushed this in another PR? no? am I reading it wrong https://github.com/Layr-Labs/eigensdk-go/pull/145/files

samlaf: Ah good call, rebased.

types/operator.go
samlaf: [d4cc682](https://github.com/Layr-Labs/eigensdk-go/pull/146/commits/d4cc68267f1dd2d04485c35350bfeaf6b72507d6)

shrimalmadhur (approved):
lgtm

PR #147, opened 2024-03-09
Update mocks

ian-shim:
Generates mock for `Wallet` and regenerates mock for `eth.Client`


chainio/gen.go
```diff
@@ -5,6 +5,7 @@ package chainio
 //go:generate mockgen -destination=./mocks/avsRegistryContractsWriter.go -package=mocks github.com/Layr-Labs/eigensdk-go/chainio/clients/avsregistry AvsRegistryWriter
 //go:generate mockgen -destination=./mocks/elContractsReader.go -package=mocks github.com/Layr-Labs/eigensdk-go/chainio/clients/elcontracts ELReader
 //go:generate mockgen -destination=./mocks/elContractsWriter.go -package=mocks github.com/Layr-Labs/eigensdk-go/chainio/clients/elcontracts ELWriter
-//go:generate mockgen -destination=./mocks/ethclient.go -package=mocks github.com/Layr-Labs/eigensdk-go/chainio/clients/eth EthClient
+//go:generate mockgen -destination=./mocks/ethclient.go -package=mocks -mock_names=Client=MockEthClient github.com/Layr-Labs/eigensdk-go/chainio/clients/eth Client
```
samlaf, line 8: the `-mock_names=Client=MockEthClient` flag makes it so that the mock name is `MockEthClient` instead of `MockClient`?

samlaf (approved):

chainio/gen.go
ian-shim: Yes. These mocks are created under a common package `mocks`, so the names need to be specific

PR #152, opened 2024-03-13
Remove signer from txn manager

ian-shim:
Removing signer from `TxManager` so that it's agnostic to the type of `wallet`. 
`TransactOpts` will be populated when the transaction is broadcasted in `TxManager.wallet`.


samlaf (approved):
@shrimalmadhur merging to unblock ian but make sure to take a look at these changes when you get a min

PR #154, opened 2024-03-14
[Hotwallet] Return sender address & set gas fee

ian-shim:
This PR adds a `SenderAddress` method to `Wallet` interface, which returns the address of the sender.
It also set a `HIGH` gas fee (configured by Fireblocks) to increase the chance of transactions being confirmed sooner.

### DIFF
--- a/chainio/clients/fireblocks/contract_call.go
+++ b/chainio/clients/fireblocks/contract_call.go
@@ -8,6 +8,7 @@ import (
 )
 
 type TransactionOperation string
+type FeeLevel string
 
 const (
 	ContractCall TransactionOperation = "CONTRACT_CALL"
@@ -16,6 +17,10 @@ const (
 	Burn         TransactionOperation = "BURN"
 	TypedMessage TransactionOperation = "TYPED_MESSAGE"
 	Raw          TransactionOperation = "RAW"
+
+	FeeLevelHigh   FeeLevel = "HIGH"
+	FeeLevelMedium FeeLevel = "MEDIUM"
+	FeeLevelLow    FeeLevel = "LOW"
 )
 
 type account struct {
@@ -38,6 +43,10 @@ type ContractCallRequest struct {
 	// In case a transaction is stuck, specify the hash of the stuck transaction to replace it
 	// by this transaction with a higher fee, or to replace it with this transaction with a zero fee and drop it from the blockchain.
 	ReplaceTxByHash string `json:"replaceTxByHash"`
+	// FeeLevel is the fee level for the transaction which Fireblocks estimates based on the current network conditions.
+	// The fee level can be HIGH, MEDIUM, or LOW.
+	FeeLevel FeeLevel `json:"feeLevel"`
+	// TODO: add maxFee and priorityFee
 }
 
 type ContractCallResponse struct {
@@ -53,6 +62,7 @@ func NewContractCallRequest(
 	amount string,
 	calldata string,
 	replaceTxByHash string,
+	feeLevel FeeLevel,
 ) *ContractCallRequest {
 	return &ContractCallRequest{
 		Operation:    ContractCall,
@@ -72,6 +82,7 @@ func NewContractCallRequest(
 			Calldata: calldata,
 		},
 		ReplaceTxByHash: replaceTxByHash,
+		FeeLevel:        feeLevel,
 	}
 }
 
--- a/chainio/clients/fireblocks/contract_call.go
+++ b/chainio/clients/fireblocks/contract_call.go
@@ -45,6 +45,7 @@ type ContractCallRequest struct {
 	ReplaceTxByHash string `json:"replaceTxByHash"`
 	// FeeLevel is the fee level for the transaction which Fireblocks estimates based on the current network conditions.
 	// The fee level can be HIGH, MEDIUM, or LOW.
+	// Ref: https://developers.fireblocks.com/docs/gas-estimation#estimated-network-fee
 	FeeLevel FeeLevel `json:"feeLevel"`
 	// TODO: add maxFee and priorityFee
 }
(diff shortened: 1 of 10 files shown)


chainio/clients/fireblocks/contract_call.go
```diff
@@ -16,6 +17,10 @@ const (
 	Burn         TransactionOperation = "BURN"
 	TypedMessage TransactionOperation = "TYPED_MESSAGE"
 	Raw          TransactionOperation = "RAW"
+
+	FeeLevelHigh   FeeLevel = "HIGH"
+	FeeLevelMedium FeeLevel = "MEDIUM"
+	FeeLevelLow    FeeLevel = "LOW"
```
samlaf, line 23: fireblocks doesn't support setting a number for gasPrice?

chainio/clients/fireblocks/get_asset_addresses.go
```diff
@@ -0,0 +1,35 @@
...
+
+func (f *client) GetAssetAddresses(ctx context.Context, vaultID string, assetID AssetID) ([]AssetAddress, error) {
+	f.logger.Debug("Fireblocks get asset addressees", "vaultID", vaultID, "assetID", assetID)
+	url := fmt.Sprintf("/v1/vault/accounts/%s/%s/addresses", vaultID, assetID)
+	res, err := f.makeRequest(ctx, "GET", url, nil)
+	if err != nil {
+		return nil, fmt.Errorf("error making request: %w", err)
+	}
+	var addresses []AssetAddress
+	err = json.NewDecoder(strings.NewReader(string(res))).Decode(&addresses)
```
samlaf, line 29: any reason you're using json.NewDecoder instead of json.Unmarshal?

chainio/clients/wallet/fireblocks_wallet.go
```diff
@@ -159,6 +159,8 @@ func (t *fireblocksWallet) SendTransaction(ctx context.Context, tx *types.Transa
 		tx.Value().String(),       // amount
 		hexutil.Encode(tx.Data()), // calldata
 		replaceTxByHash,           // replaceTxByHash
+		// TODO: make this configurable
+		fireblocks.FeeLevelHigh, // feeLevel
```
samlaf, line 163: why not pass it as argument to SendTransaction?

chainio/clients/wallet/wallet.go
```diff
@@ -16,4 +17,5 @@ type TxID = string
 type Wallet interface {
 	SendTransaction(ctx context.Context, tx *types.Transaction) (TxID, error)
 	GetTransactionReceipt(ctx context.Context, txID TxID) (*types.Receipt, error)
+	SenderAddress(ctx context.Context) (common.Address, error)
```
samlaf, line 20: Why is this needed? Can you add comment explaining its usecase. Doesn't seem like its used anywhere in this PR, unless I missed it.

chainio/clients/fireblocks/contract_call.go
ian-shim: It does, and we should add them too in the future, but I wanted to just use the fees Fireblocks estimates for us given the fee level

chainio/clients/fireblocks/get_asset_addresses.go
ian-shim: Not exactly sure what the difference is. It was what fireblocks example code was using 😆

chainio/clients/wallet/fireblocks_wallet.go
ian-shim: `SendTransaction` should share a common signature for all wallets. `feeLevel` field is specific to Fireblocks wallet.

chainio/clients/wallet/wallet.go
ian-shim: I wanted to use this when you pass in the sender address when you create the txMgr

chainio/clients/fireblocks/contract_call.go
samlaf: sgtm. Can you just add a TODO comment then to support those, unless its already written somewhere else?

chainio/clients/fireblocks/get_asset_addresses.go
samlaf: haha sounds good, was just curious

chainio/clients/wallet/fireblocks_wallet.go
samlaf: hmm I see. so when/how would you set the fee level then? Say if we calculate a gasPrice or whatever? Also what does adding this feeLevel do? Is fireblocks going to replace the gasPrice set in the tx??

chainio/clients/wallet/wallet.go
samlaf: yep ok just saw this in avs-sync! Ok that makes sense. Can you just add a comment here explaining that this is needed to create a txMgr that wraps the wallet, since the wallet hides the signer key/address, whereas the txMgr needs to know the from address (forget why though lol)

chainio/clients/wallet/fireblocks_wallet.go
ian-shim: Right now, it sends every transaction with `HIGH` gas. There is a TODO for making it configurable. 
If you look at how `SendTransaction` works in this method, it only uses `To`, `Hash`, `Value`, and `Data` of the input transaction. So Fireblocks isn't aware of the gas specified in the `tx` object unless it's passed in in the API call. Right now, it's ignoring the gas fees set in the `tx` and always sending them with `HIGH` fee.

chainio/clients/wallet/wallet.go
ian-shim: done

chainio/clients/fireblocks/contract_call.go
ian-shim: done

chainio/clients/wallet/fireblocks_wallet.go
samlaf: gotcha, thanks for the explanation.

samlaf (approved):

PR #155, opened 2024-03-16
Few fixes on `Wallet` implementation

ian-shim:
- Renamed `Status` with `TxStatus` to make it clear that it only applies to transaction status
- Fixed a bug in fireblocks wallet `getWhitelistedContract` method where it was returning the last contract instead of the contract that matches the condition
- Fixed a bug in txmgr `queryReceipt` method which was transforming `txID` into a hash (this only worked for private key wallet)


samlaf (approved):
LGTM. I'll trust you on the fireblocks changes given you've tested them with avs-sync.

PR #159, opened 2024-03-22
[Fireblocks] Don't replace transaction if txHash not available

ian-shim:
`fireblockTx.TxHash` isn't available if the transaction has not been broadcasted to the network. We don't need to specify `replaceTxByHash` in this case. 
Screenshot 2024-03-22 at 1 36 58 PM

### DIFF
--- a/chainio/clients/wallet/fireblocks_wallet.go
+++ b/chainio/clients/wallet/fireblocks_wallet.go
@@ -152,8 +152,6 @@ func (t *fireblocksWallet) SendTransaction(ctx context.Context, tx *types.Transa
 		}
 		if fireblockTx.TxHash != "" {
 			replaceTxByHash = fireblockTx.TxHash
-		} else {
-			return "", fmt.Errorf("failed to get transaction hash with nonce %d", nonce)
 		}
 	}
 


mooselumph:
I'm confused about how the transaction would be replaced in this case. The nonce isn't passed on, so how does it know to replace the transaction instead of creating a new one?

ian-shim:
> I'm confused about how the transaction would be replaced in this case. The nonce isn't passed on, so how does it know to replace the transaction instead of creating a new one?

Fireblocks does not take nonce as a parameter for `CallContract` call. Instead, it lets you specify which transaction to replace by passing in `replaceTxByHash`. We have to keep track of nonces and corresponding tx hashes to make this interface compatible with geth client.

mooselumph:
> Fireblocks does not take nonce as a parameter for `CallContract` call. Instead, it lets you specify which transaction to replace by passing in `replaceTxByHash`. We have to keep track of nonces and corresponding tx hashes to make this interface compatible with geth client.

So just to be clear, this will result in a new transaction instead of the old one being replaced when that was the intent? Isn't this an issue in that when the intent was to speed up a transaction, instead a new one will be created?

ian-shim:
> > Fireblocks does not take nonce as a parameter for `CallContract` call. Instead, it lets you specify which transaction to replace by passing in `replaceTxByHash`. We have to keep track of nonces and corresponding tx hashes to make this interface compatible with geth client.
> 
> So just to be clear, this will result in a new transaction instead of the old one being replaced when that was the intent? Isn't this an issue in that when the intent was to speed up a transaction, instead a new one will be created?

Yes, this will result in a new transaction. Not having txHash back means that the previous transaction with the same nonce was not broadcasted to the network properly though, so I think it does make sense to send a new transaction

jianoaix (approved):

PR #160, opened 2024-03-23
[Fireblocks] Pass in gas info when calling contract

ian-shim:
It used to rely on Fireblocks' gas estimation by passing in `HIGH` as `feeLevel` parameter when creating a transaction.
However, this estimation doesn't guarantee that replacement (boosting) transaction increases the gas cost by > 15% than the previous transaction, which is a requirement for replacing a pending transaction. 
We need to explicitly pass in the gas fee cap and priority fee cap instead of passing in `feeLevel` in this case.

### DIFF
--- a/chainio/clients/fireblocks/contract_call.go
+++ b/chainio/clients/fireblocks/contract_call.go
@@ -43,11 +43,20 @@ type ContractCallRequest struct {
 	// In case a transaction is stuck, specify the hash of the stuck transaction to replace it
 	// by this transaction with a higher fee, or to replace it with this transaction with a zero fee and drop it from the blockchain.
 	ReplaceTxByHash string `json:"replaceTxByHash"`
+	// GasPrice and GasLimit are the gas price and gas limit for the transaction.
+	// If GasPrice is specified (non-1559), MaxFee and PriorityFee are not required.
+	GasPrice string `json:"gasPrice"`
+	GasLimit string `json:"gasLimit"`
+	// MaxFee and PriorityFee are the maximum and priority fees for the transaction.
+	// If the transaction is stuck, the Fireblocks platform will replace the transaction with a new one with a higher fee.
+	// These fields are required if FeeLevel is not specified.
+	MaxFee      string `json:"maxFee"`
+	PriorityFee string `json:"priorityFee"`
 	// FeeLevel is the fee level for the transaction which Fireblocks estimates based on the current network conditions.
 	// The fee level can be HIGH, MEDIUM, or LOW.
+	// If MaxFee and PriorityFee are not specified, the Fireblocks platform will use the default fee level MEDIUM.
 	// Ref: https://developers.fireblocks.com/docs/gas-estimation#estimated-network-fee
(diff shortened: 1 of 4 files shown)


chainio/clients/fireblocks/contract_call.go
```diff
@@ -43,11 +43,17 @@ type ContractCallRequest struct {
 	// In case a transaction is stuck, specify the hash of the stuck transaction to replace it
 	// by this transaction with a higher fee, or to replace it with this transaction with a zero fee and drop it from the blockchain.
 	ReplaceTxByHash string `json:"replaceTxByHash"`
+	gasLimit        string `json:"gasLimit"`
+	// MaxFee and PriorityFee are the maximum and priority fees for the transaction.
+	// If the transaction is stuck, the Fireblocks platform will replace the transaction with a new one with a higher fee.
+	// These fields are required if FeeLevel is not specified.
+	MaxFee      string `json:"maxFee"`
+	PriorityFee string `json:"priorityFee"`
```
samlaf, line 51: shouldn't this be MaxPriorityFee? See eip1559 https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1559.md

See also https://ethereum.stackexchange.com/a/134950/85362 for why this needs to be max and not fixed.

chainio/clients/wallet/fireblocks_wallet.go
```diff
@@ -155,6 +156,23 @@ func (t *fireblocksWallet) SendTransaction(ctx context.Context, tx *types.Transa
...
+	}
+
+	// if the gas fees are specified in the transaction, use them.
+	// Otherwise, use the default "HIGH" gas price estimated by Fireblocks
+	maxFee := ""
+	priorityFee := ""
+	feeLevel := fireblocks.FeeLevel("")
+	if tx.GasFeeCap().Cmp(big.NewInt(0)) > 0 && tx.GasTipCap().Cmp(big.NewInt(0)) > 0 {
+		maxFee = tx.GasFeeCap().String()
+		priorityFee = tx.GasTipCap().String()
```
samlaf, line 171: Do we also need to support legacy txs for whatever reason?
See in the eip
image

chainio/clients/fireblocks/contract_call.go
ian-shim: Note that these fields are being used to call a Fireblocks API endpoint, not ETH client.
So the field names don't exactly match those defined on ETH client: https://developers.fireblocks.com/reference/post_transactions
The `priorityFee` field matches what you call maxPriorityFee.
Afaik, there's not even a way to set just "priority fee" on a transaction

chainio/clients/wallet/fireblocks_wallet.go
ian-shim: Sure, callers would usually construct an 1559 transaction. But I added `gasPrice` field in case the caller wants to construct a non-1559 transaction.

samlaf (approved):
lgtm

PR #162, opened 2024-03-27
[logger] Add `With` method for creating child logger

ian-shim:
`With` method can be used to create child logger with default tags


samlaf (approved):
lgtm

PR #163, opened 2024-03-27
[fireblocks] omit empty `CallContract` fields

ian-shim:
For Fireblocks `CallContract` method, it rejects requests if multiple gas fee mechanisms are specified. A request should only specify one of:
- `gasPrice` (non-1559)
- `maxFee` and `priorityFee` (1559)
- `feeLevel` (Fireblocks gas estimation)

This PR adds `omitempty` tag to optional fields so that these fields are omitted in the json request body.


chainio/clients/fireblocks/contract_call.go
```diff
@@ -96,12 +96,19 @@ func NewContractCallRequest(
...
-		FeeLevel:        feeLevel,
 	}
+
+	if maxFee != "" && priorityFee != "" {
+		req.MaxFee = maxFee
+		req.PriorityFee = priorityFee
+	} else if gasPrice != "" {
+		req.GasPrice = gasPrice
+	} else {
+		req.FeeLevel = feeLevel
```
jianoaix, line 108: can `feeLevel` passed in from NewContractCallRequest be empty?

chainio/clients/fireblocks/contract_call.go
```diff
@@ -96,12 +96,19 @@ func NewContractCallRequest(
 			Calldata: calldata,
 		},
 		ReplaceTxByHash: replaceTxByHash,
-		GasPrice:        gasPrice,
 		GasLimit:        gasLimit,
-		MaxFee:          maxFee,
-		PriorityFee:     priorityFee,
-		FeeLevel:        feeLevel,
 	}
+
```
jianoaix, line 101: Is it caller (of NewContractCallRequest) responsibility to ensure exactly one of the 3 options is set?
In that case it will make sense to validate, or at least log if this assumption doesn't hold

ian-shim: The caller doesn't necessarily have to specify any of the options above, in which case Fireblocks will default `feeLevel` to `MEDIUM`

chainio/clients/fireblocks/contract_call.go
ian-shim: Yes, in which case Fireblocks will default it to MEDIUM

jianoaix (approved):

chainio/clients/fireblocks/contract_call.go
jianoaix: Might be helpful to leave a comment here about it's handled this way

samlaf: it's specified on line 57

PR #168, opened 2024-03-29
[fireblocks] Convert gas units to gwei

ian-shim:
Apparently, Fireblocks expects gas fees to be in gwei (instead of wei)


samlaf (approved):
lgtm

PR #171, opened 2024-03-31
[fireblocks] Adjust confirmation threshold

ian-shim:
Allow setting confirmation threshold for fireblocks transactions (which default to 3 otherwise)


mooselumph:
Seems like this PR sets a fixed confirmation threshold for the wallet and then updates the confirmation threshold for each individual transaction. I suppose there isn't a global fireblocks setting that can be used instead?

ian-shim:
> Seems like this PR sets a fixed confirmation threshold for the wallet and then updates the confirmation threshold for each individual transaction. I suppose there isn't a global fireblocks setting that can be used instead?

There doesn't seem like an option to set the global threshold for the account 😢

chainio/clients/wallet/fireblocks_wallet.go
```diff
@@ -199,6 +201,14 @@ func (t *fireblocksWallet) SendTransaction(ctx context.Context, tx *types.Transa
 	t.txIDToNonce[res.ID] = nonce
 	t.logger.Debug("Fireblocks contract call complete", "txID", res.ID, "status", res.Status)
 
+	success, err := t.fireblocksClient.SetConfirmationThreshold(ctx, res.ID, t.confirmationThreshold)
```
mooselumph, line 204: Maybe use `confirmationThreshold=0` or `-1` to just use the default?

mooselumph (approved):
lgtm

PR #177, opened 2024-04-04
Fix log label from `tx hash` to `txHash`

ian-shim:
Removing space from log attribute labels


samlaf (approved):
Curious, did this create a problem for loki/grafana, or is this just for aesthetic reasons?

Also in general, what do you think of the verbosity of our logging? I'm starting to find it a bit too verbose. But also the problem is user of the sdk doesn't have ability to turn these logs off. Was thinking either to remove them, or at least make them debug logs. Or otherwise maybe have a `verbose` arg when building the different clients, that only logs when turned on?

ian-shim:
I was having trouble filtering it in Loki, and I think it's good idea to not have spaces in labels in general. 
Yeah it could get noisy, but I think log level should serve that purpose well. Wouldn't providing a high log level be equivalent to non-verbose mode?

PR #179, opened 2024-04-04
Fix:  Error wrap may panic by nil error

fyInALT:
### Motivation

Note in:

```go
func (ar *AvsRegistryServiceChainCaller) getOperatorPubkeys(ctx context.Context, operatorId types.OperatorId) (types.OperatorPubkeys, error) {
	operatorAddr, err := ar.AvsRegistryReader.GetOperatorFromId(&bind.CallOpts{Context: ctx}, operatorId)
	if err != nil {
		return types.OperatorPubkeys{}, types.WrapError(errors.New("Failed to get operator address from pubkey hash"), err)
	}
	pubkeys, ok := ar.operatorPubkeysService.GetOperatorPubkeys(ctx, operatorAddr)
	if !ok {
		return types.OperatorPubkeys{}, fmt.Errorf("Failed to get operator pubkeys from pubkey compendium service (operatorAddr: %v, operatorId: %v)", operatorAddr, operatorId)
	}
	return pubkeys, nil
}

```

if `operatorPubkeysService` not found the key, it will return false but the `err` is nil, it will cause a panic by read nil point by wrapper:

```go
fmt.Errorf("%s: %s", mainErr.Error(), subErr.Error())
```

### Solution

- we need add checks in wrapper, even use nil error, will not cause a panic
- fix the operatorPubkeysService error wrapper, the wrap is useless by the err will always nil

### DIFF
--- a/services/avsregistry/avsregistry_chaincaller.go
+++ b/services/avsregistry/avsregistry_chaincaller.go
@@ -109,7 +109,7 @@ func (ar *AvsRegistryServiceChainCaller) getOperatorPubkeys(ctx context.Context,
 	}
 	pubkeys, ok := ar.operatorPubkeysService.GetOperatorPubkeys(ctx, operatorAddr)
 	if !ok {
-		return types.OperatorPubkeys{}, types.WrapError(fmt.Errorf("Failed to get operator pubkeys from pubkey compendium service (operatorAddr: %v, operatorId: %v)", operatorAddr, operatorId), err)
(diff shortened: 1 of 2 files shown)


samlaf (approved):
LGTM. Been thinking we should just use https://pkg.go.dev/errors#Join instead of our own WrapError function. That one takes a vararg list of errors and just filters out the nil errors passed. But it prints using '\n' separators between the errors which some people don't like... not sure what's the best way here, but at least thanks for fixing these errors. :)

PR #181, opened 2024-04-04
better twitter validation

shrimalmadhur:
Fixes # .

### Motivation
- Add more strict `twitter` and `x` validations

### DIFF
--- a/types/operator_metadata.go
+++ b/types/operator_metadata.go
@@ -93,7 +93,7 @@ func checkIfValidTwitterURL(twitterURL string) error {
 	}
 
 	// Regular expression to validate URLs
-	urlPattern := regexp.MustCompile(`^(?:https?://)?(?:www\.)?twitter\.com/\w+|x\.com/\w+$`)
+	urlPattern := regexp.MustCompile(`^(?:https?://)?(?:www\.)?(?:twitter\.com/\w+|x\.com/\w+)(?:/?|$)`)
 
 	// Check if the URL matches the regular expression
 	if !urlPattern.MatchString(twitterURL) {
--- a/types/operator_metadata.go
+++ b/types/operator_metadata.go
@@ -76,7 +76,7 @@ func (om *OperatorMetadata) Validate() error {
 	}
 
 	if len(om.Twitter) != 0 {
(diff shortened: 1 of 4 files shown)


joan-el (dismissed):
this looks good, wait for Anto for final blessing

types/operator_metadata_test.go
```diff
@@ -22,6 +22,16 @@ func TestOperatorMetadata(t *testing.T) {
 				Website:     "https://test.com",
 			},
 		},
+		{
```
joan-el, line 25: love it you added this. as well

types/operator_metadata.go
```diff
@@ -85,7 +85,25 @@ func (om *OperatorMetadata) Validate() error {
...
-func checkIfUrlIsValid(rawUrl string) error {
+func checkIfValidTwitterURL(twitterURL string) error {
+	// Basic validation
+	err := checkBasicURLValidation(twitterURL)
+	if err != nil {
+		return err
+	}
+
+	// Regular expression to validate URLs
+	urlPattern := regexp.MustCompile(`^(?:https?://)?(?:www\.)?twitter\.com/\w+|x\.com/\w+$`)
```
antojoseph, line 96: Does this make sense?
`^(?:https?://)?(?:www\.)?(?:twitter\.com/\w+|x\.com/\w+)(?:/?|$)
`

antojoseph (commented):
added a regex suggestion, please check

types/operator_metadata.go
shrimalmadhur: why would twitter have `:` in it @antojoseph ? can you give an example of what url you are thinking?

shrimalmadhur: oh I se. lol I think i read it wrong

antojoseph: we can remove that, i don't think ppl specify their twitter links like https://twitter.com:443/abcd

shrimalmadhur: It makes sense

shrimalmadhur: updated

antojoseph (approved):
lgtm

PR #183, opened 2024-04-04
validate urls

shrimalmadhur:
Fixes # .

### Motivation
- Add checks so that there's no redirect
- Add Github validation which should be used by cli and webserver
- Bunch of refactor

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -4,11 +4,16 @@ import (
 	"crypto/ecdsa"
 	"encoding/json"
 	"errors"
+
 	gethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/crypto"
 	"gopkg.in/yaml.v3"
 	"log"
 	"math/big"
+
+	"fmt"
+	"io"
+	"net/http"
 	"os"
 	"path/filepath"
 	"regexp"
@@ -74,4 +79,30 @@ func RoundUpDivideBig(a, b *big.Int) *big.Int {
 func IsValidEthereumAddress(address string) bool {
 	re := regexp.MustCompile("^0x[0-9a-fA-F]{40}$")
 	return re.MatchString(address)
+}
+
+func ReadPublicUrl(url string) ([]byte, error) {
+	// allow no redirects
+	httpClient := http.Client{
+		CheckRedirect: func(req *http.Request, via []*http.Request) error {
+			return http.ErrUseLastResponse
+		},
+	}
+	resp, err := httpClient.Get(url)
+	if err != nil {
+		return []byte{}, err
+	}
+
+	if resp.StatusCode >= 400 {
+		return []byte{}, fmt.Errorf("error fetching url: %s", resp.Status)
+	}
+
+	defer func(Body io.ReadCloser) {
+		err := Body.Close()
+		if err != nil {
+			fmt.Println("error closing url body")
+		}
+	}(resp.Body)
+
+	return io.ReadAll(resp.Body)
 }
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -4,6 +4,8 @@ import (
 	"crypto/ecdsa"
 	"encoding/json"
 	"errors"
+	"net/url"
+	"strings"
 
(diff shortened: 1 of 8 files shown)


utils/utils.go
```diff
@@ -75,3 +94,160 @@ func IsValidEthereumAddress(address string) bool {
...
+	textPattern := regexp.MustCompile(`^[a-zA-Z0-9 .,;:?!'"\-_/()\[\]]+$`)
+
+	// Check if the URL matches the regular expression
+	if !textPattern.MatchString(text) {
+		return ErrInvalidText
+	}
+
+	return nil
+}
+func ValidateRawGithubUrl(url string) error {
```
shrimalmadhur, line 237: @antojoseph I added this here but have not used directly in eigensdk. Reason being we only want to use this validation on mainnet so I will add that logic in CLI and server. wdyt?

utils/utils.go
```diff
@@ -75,3 +94,160 @@ func IsValidEthereumAddress(address string) bool {
 	re := regexp.MustCompile("^0x[0-9a-fA-F]{40}$")
 	return re.MatchString(address)
 }
+
+func ReadPublicUrl(url string) ([]byte, error) {
```
shrimalmadhur, line 98: @antojoseph this is what you should review

utils/utils.go
antojoseph: Yeah, Adding it to the server sounds good to me.

utils/utils.go
```diff
@@ -75,3 +95,161 @@ func IsValidEthereumAddress(address string) bool {
...
+}
+func ValidateRawGithubUrl(url string) error {
+	// Basic validation
+	err := CheckBasicURLValidation(url)
+	if err != nil {
+		return err
+	}
+
+	// Regular expression to validate URLs
+	rawGitHubUrlPattern := regexp.MustCompile(`^https?://raw\.githubusercontent\.com/.*$`)
```
antojoseph, line 247: Looks good to me, maybe we can only allow https?

utils/utils.go
```diff
@@ -75,3 +95,161 @@ func IsValidEthereumAddress(address string) bool {
...
+	parsedURL, err := url.Parse(urlString)
+	if err != nil {
+		return err
+	}
+
+	// Extract the path component from the URL
+	path := parsedURL.Path
+
+	// Get the file extension
+	extension := filepath.Ext(path)
```
antojoseph, line 199: tested with files like malware.sh.png, it will return png , although the file type maynot be png

utils/utils.go
```diff
@@ -75,3 +95,161 @@ func IsValidEthereumAddress(address string) bool {
...
+
+	// Check if the extension is in the list of image extensions
+	for _, imgExt := range ImageExtensions {
+		if strings.EqualFold(extension, imgExt) {
+			imageBytes, err := ReadPublicUrl(urlString)
+			if err != nil {
+				return err
+			}
+
+			contentType := http.DetectContentType(imageBytes)
```
antojoseph, line 209: Doing this check solves uploading files like malware.sh.png

utils/utils.go
```diff
@@ -75,3 +95,161 @@ func IsValidEthereumAddress(address string) bool {
 	re := regexp.MustCompile("^0x[0-9a-fA-F]{40}$")
 	return re.MatchString(address)
 }
+
+func ReadPublicUrl(url string) ([]byte, error) {
+	// allow no redirects
+	httpClient := http.Client{
```
antojoseph, line 101: This stops the client from following re-directs

antojoseph (approved):
lgtm

utils/utils.go
shrimalmadhur: yea the mimetype check will fail - do you want any stricter check?

utils/utils.go
shrimalmadhur: yeah!

utils/utils.go
shrimalmadhur: I see - can do a fast follow

Issue #187, opened 2024-04-08
blsagg service doesn't returned sorted operator pubkeys


samlaf:
**Describe the bug**

checkSignatures [expects nonsigner operator pubkeys to be sorted](https://github.com/Layr-Labs/eigenlayer-middleware/blob/dev/src/BLSSignatureChecker.sol#L149), but the blsagg service currently doesn't sort operator pubkeys. See https://github.com/Layr-Labs/eigensdk-go/blob/977516719b1a7b6ec00901175f49e0ead6c4917c/services/bls_aggregation/blsagg.go#L403

**To Reproduce**
Don't have a reproducible test right now. This was raised by an avs team. Will need to add a test case for this.

i9:
thanks! we hope this will also fixed undetermined order of nonSignersOperatorIds and NonSignersPubkeysG1 (due to golang map iter random order)

samlaf:
Fixed in #188

PR #188, opened 2024-04-08
fix issue #187

hezhihua81:


services/bls_aggregation/blsagg.go
```diff
@@ -314,6 +326,43 @@ func (a *BlsAggregatorService) singleTaskAggregatorGoroutineFunc(
 
 }
 
+type Wrapper struct {
```
samlaf, line 329: can we make this private? `type wrapper struct {...}`

services/bls_aggregation/blsagg.go
```diff
@@ -301,6 +305,14 @@ func (a *BlsAggregatorService) singleTaskAggregatorGoroutineFunc(
...
 					NonSignerStakeIndices:        indices.NonSignerStakeIndices,
 				}
+
+				blsAggregationServiceResponse.NonSignersPubkeysG1,
+					blsAggregationServiceResponse.NonSignerQuorumBitmapIndices =
+					sortByOprId(
+						blsAggregationServiceResponse.NonSignersPubkeysG1,
+						blsAggregationServiceResponse.NonSignerQuorumBitmapIndices,
+					)
+
```
samlaf, line 315: I think that this needs to be moved above, before the call to GetCheckSignaturesIndices, otherwise the NonSignerStakeIndices returned won't be in the correct order?

If this worked for you, I'm a bit puzzled. Am I missing something here?

samlaf (changes requested):
Thanks a bunch for finding and fixing this! One minor and one major comment to address, then we'll be good to merge.

services/bls_aggregation/blsagg.go
hezhihua81: refined

services/bls_aggregation/blsagg.go
hezhihua81: refined

samlaf (approved):
Beautiful. Thanks for this :)

PR #195, opened 2024-04-09
[fireblocks] Return specific err for prebroadcast status

ian-shim:
Adds `ErrNotYetBroadcasted` to indicate that the transaction is still being processed by Fireblocks without having been published in the network.
Callers can use this information to customize logic that waits for the transaction receipt.

### DIFF
--- a/chainio/clients/wallet/fireblocks_wallet_test.go
+++ b/chainio/clients/wallet/fireblocks_wallet_test.go
@@ -323,7 +323,30 @@ func TestWaitForTransactionReceiptFailFromFireblocks(t *testing.T) {
 	}, nil)
 
 	receipt, err := sender.GetTransactionReceipt(context.Background(), expectedTxHash)
-	assert.Error(t, err)
+	assert.ErrorAs(t, err, &wallet.ErrReceiptNotYetAvailable)
+	assert.Nil(t, receipt)
+}
+
+func TestWaitForTransactionReceiptStuckAtFireblocks(t *testing.T) {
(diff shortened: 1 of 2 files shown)


chainio/clients/wallet/fireblocks_wallet_test.go
```diff
@@ -323,7 +323,30 @@ func TestWaitForTransactionReceiptFailFromFireblocks(t *testing.T) {
...
+	assert.ErrorAs(t, err, &wallet.ErrReceiptNotYetAvailable)
+	assert.Nil(t, receipt)
+}
+
+func TestWaitForTransactionReceiptStuckAtFireblocks(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+	fireblocksClient := cmocks.NewMockFireblocksClient(ctrl)
+	ethClient := mocks.NewMockEthClient(ctrl)
+	logger, err := logging.NewZapLogger(logging.Development)
```
samlaf, line 335: just a nit as this is just a test, but let's try to only use slog moving forward, to be consistent inside the sdk.

samlaf (approved):

chainio/clients/wallet/fireblocks_wallet_test.go
ian-shim: Agree we should try to use one type of logger in the codebase. Are we going with slog? I don't see any slog usage in this repo, but do see zap logger being used

PR #198, opened 2024-04-09
[fireblocks] Add a way to cancel transaction

ian-shim:
Adding support for canceling transaction in flight. 
This method can cancel transactions that have not been broadcasted to the network (either because it's still gathering cosignatures or other MPC operations)

### DIFF
--- a/chainio/clients/fireblocks/cancel_transaction.go
+++ b/chainio/clients/fireblocks/cancel_transaction.go
@@ -1,0 +1,28 @@
+package fireblocks
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"strings"
+)
+
+type CancelTransactionResponse struct {
+	Success bool `json:"success"`
+}
+
+func (f *client) CancelTransaction(ctx context.Context, txID string) (bool, error) {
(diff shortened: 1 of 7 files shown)


chainio/clients/fireblocks/cancel_transaction.go
```diff
@@ -0,0 +1,28 @@
...
+	"strings"
+)
+
+type CancelTransactionResponse struct {
+	Success bool `json:"success"`
+}
+
+func (f *client) CancelTransaction(ctx context.Context, txID string) (bool, error) {
+	f.logger.Debug("Fireblocks cancel transaction", "txID", txID)
+	url := fmt.Sprintf("/v1/transactions/%s/cancel", txID)
```
samlaf, line 16: nit: this is a path, not a url

samlaf (approved):

chainio/clients/fireblocks/cancel_transaction.go
ian-shim: updated

samlaf (approved):
LGTM! sorry, this slipped in my mailbox

PR #203, opened 2024-04-16
Refactor/wrap error type

### LINKED ISSUE #116: Move WrapError to utils

- Currently WrapError function lives in `types` which doesn't seem right place
- Interface expects two errors which is not great since, a lot of places has a string input (and error inputs in some place)

- Move WrapError to utils or similar
- Interface should accept `any` type and then based on type do the Wrapping.

**Additional context**

renlulu:
Fixes #116  .

### DIFF
--- a/utils/errors.go
+++ b/utils/errors.go
@@ -52,5 +52,5 @@ func WrapError(mainErr interface{}, subErr interface{}) error {
 		return main
 	}
 
-	return fmt.Errorf("%v: %v", mainErr, subErr)
+	return fmt.Errorf("%w: %w", main, sub)
 }
--- a/utils/errors.go
+++ b/utils/errors.go
@@ -52,5 +52,5 @@ func WrapError(mainErr interface{}, subErr interface{}) error {
 		return main
 	}
 
-	return main
+	return fmt.Errorf("%v: %v", mainErr, subErr)
 }
--- a/utils/errors.go
+++ b/utils/errors.go
@@ -23,3 +23,34 @@ var (
 		"invalid twitter url, it should be of the format https://twitter.com/<username> or https://x.com/<username>",
 	)
 )
+
+func TypedErr(e interface{}) error {
+	switch t := e.(type) {
+	case error:
+		return t
+	case string:
+		return errors.New(t)
+	default:
+		return nil
+	}
+}
+
+func WrapError(mainErr interface{}, subErr interface{}) error {
+	var main, sub error
+	main = TypedErr(mainErr)
+	sub = TypedErr(subErr)
+	// Some times the wrap will wrap a nil error
+	if main == nil && sub == nil {
+		return nil
+	}
+
+	if main == nil && sub != nil {
+		return sub
+	}
+
+	if main != nil && sub == nil {
+		return main
+	}
+
+	return main
+}
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -279,12 +279,12 @@ func WrapError(mainErr interface{}, subErr interface{}) error {
 	}
 
 	if main == nil && sub != nil {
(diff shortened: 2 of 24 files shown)


samlaf:
@renlulu sorry took us a while to review this, looks good. Just merged another PR which created merge conflict.. really sorry.
Would you mind rebasing on top of master? Then I'll review and we can merge promptly.

renlulu:
> @renlulu sorry took us a while to review this, looks good. Just merged another PR which created merge conflict.. really sorry. Would you mind rebasing on top of master? Then I'll review and we can merge promptly.

Yes resolving!

renlulu:
@samlaf Hey can you confirm the checking error? Looks like there is some permission issue?

samlaf:
> @samlaf Hey can you confirm the checking error? Looks like there is some permission issue?

That's unfortunately expected... the way we upload the badges on the README requires committer to be in eigenlabs org right now.. we have an issue to eventually fix this, but we can merge with this failing.

utils/utils.go
```diff
@@ -256,3 +257,34 @@ func ValidateRawGithubUrl(url string) error {
...
+	if main == nil && sub == nil {
+		return nil
+	}
+
+	if main == nil && sub != nil {
+		return fmt.Errorf("sub error: %w", sub)
+	}
+
+	if main != nil && sub == nil {
+		return fmt.Errorf("%w: unknown sub error", main)
```
samlaf, line 286: how about just returning main instead?
this follows the [errors.Join](https://pkg.go.dev/errors#Join) semantic where they just discard any nil error argument

utils/utils.go
```diff
@@ -256,3 +257,34 @@ func ValidateRawGithubUrl(url string) error {
...
+	var main, sub error
+	main = TypedErr(mainErr)
+	sub = TypedErr(subErr)
+	// Some times the wrap will wrap a nil error
+	if main == nil && sub == nil {
+		return nil
+	}
+
+	if main == nil && sub != nil {
+		return fmt.Errorf("sub error: %w", sub)
```
samlaf, line 282: is this needed? why not just return sub?

utils/utils.go
```diff
@@ -256,3 +257,34 @@ func ValidateRawGithubUrl(url string) error {
...
+	case string:
+		return errors.New(t)
+	default:
+		return nil
+	}
+}
+
+func WrapError(mainErr interface{}, subErr interface{}) error {
+	var main, sub error
+	main = TypedErr(mainErr)
```
samlaf, line 274: can you move all of these to utils/errors.go instead?

samlaf (changes requested):

utils/utils.go
renlulu: Okie, make sense. I was just following the initial returned value style. Changing now.

utils/utils.go
renlulu: > Just done!

utils/errors.go
```diff
@@ -23,3 +23,34 @@ var (
...
+
+	if main == nil && sub != nil {
+		return sub
+	}
+
+	if main != nil && sub == nil {
+		return main
+	}
+
+	return main
```
samlaf, line 55: you're not wrapping anymore.

samlaf (changes requested):

utils/errors.go
samlaf:
```suggestion
	return fmt.Errorf("%w: %w", main, sub)
```

renlulu: on ok!

samlaf (approved):
LGTM. reran the flaky test and it passed. Will solve in a separate PR. Let's merge this. :)

PR #206, opened 2024-04-19
Add socket info to operatorPubkeyService (also renamed OperatorInfoService)

samlaf:
This is actually a pretty simple change that has a lot of downstream effects.
Need socket info registered onchain from operators to be returned by the operator service.
Had to also update upstream services that depend on this service.


types/operator.go
```diff
@@ -62,14 +64,30 @@ func (o Operator) Validate() error {
 	return operatorMetadata.Validate()
 }
 
-// Should we also add the operator address in here?
+type Socket = string
```
shrimalmadhur, line 67: change this to type and remove aliases

samlaf: [8f1d3f1](https://github.com/Layr-Labs/eigensdk-go/pull/206/commits/8f1d3f1171eb2f97b4e749ab98c91e4db0650923)

shrimalmadhur (approved):

PR #212, opened 2024-04-23
Fix operatorsinfo socket bugs

samlaf:
socket event filter iterator logic was totally wrong so was returning 0 results. Fixed it here.
Also added some other cosmetic changes.


services/operatorsinfo/operatorsinfo_inmemory.go
```diff
@@ -189,3 +189,10 @@ func (ops *OperatorsInfoServiceInMemory) GetOperatorInfo(ctx context.Context, op
 		return resp.operatorInfo, resp.operatorExists
 	}
 }
+
+func (ops *OperatorsInfoServiceInMemory) updateSocketMapping(operatorId types.OperatorId, socket types.Socket) {
+	if socket == "" {
+		ops.logger.Warn("Received empty socket for operator", "operatorId", operatorId)
```
shrimalmadhur, line 195: add a return here since it's a no-op if socket is empty

samlaf: [0aefb65](https://github.com/Layr-Labs/eigensdk-go/pull/212/commits/0aefb653a83afe3506107486519b61a5f60e01d3)

shrimalmadhur (approved):

PR #217, opened 2024-04-24
[fireblocks] Leave `externalTxID` empty

ian-shim:
`externalTxID` serves as an idempotency key.
An issue with Fireblocks `externalTxID` is that even when the `ContractCall` request fails (i.e. due to timeout) their DB still saves the `externalTxID`, which invalidates any further retries with the same `externalTxID`. With the information available in `types.Transaction`, there isn't reliable way to determine if the request is a retry or not.

### DIFF
--- a/chainio/clients/fireblocks/contract_call.go
+++ b/chainio/clients/fireblocks/contract_call.go
@@ -33,13 +33,14 @@ type extraParams struct {
 }
 
 type ContractCallRequest struct {
-	Operation       TransactionOperation `json:"operation"`
-	ExternalTxID    string               `json:"externalTxId"`
-	AssetID         AssetID              `json:"assetId"`
-	Source          account              `json:"source"`
-	Destination     account              `json:"destination"`
-	Amount          string               `json:"amount,omitempty"`
-	ExtraParameters extraParams          `json:"extraParameters"`
+	Operation TransactionOperation `json:"operation"`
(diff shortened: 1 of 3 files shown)


chainio/clients/fireblocks/contract_call.go
```diff
@@ -34,7 +34,7 @@ type extraParams struct {
 
 type ContractCallRequest struct {
 	Operation       TransactionOperation `json:"operation"`
-	ExternalTxID    string               `json:"externalTxId"`
+	ExternalTxID    string               `json:"externalTxId,omitempty"`
```
samlaf, line 37: add the explanation you made in the PR definition here as a comment, so we know why omitempty is fine

samlaf (commented):
LGTM. tests are failing though

chainio/clients/fireblocks/contract_call.go
ian-shim: 👍 clarified that this is an optional field

samlaf (approved):
LGTM. Out of curiosity, when would one want to set ExternalTxID then?

ian-shim:
> LGTM. Out of curiosity, when would one want to set ExternalTxID then?

That's a good question. I initially added it because Fireblocks recommends setting `ExternalTxID`. You'd want one if it's possible that the system can submit multiple requests for the same contract call and you need only one request be submitted no matter if the request times out or not

PR #221, opened 2024-04-29
fix: duplicate operatorAddresses, operatorPubkeys caused by inclusive FilterQuery

taco-paco:
Fixes # .

### What Changed?

Since FilterQuery is inclusive (as discussed [here](https://ethereum.stackexchange.com/questions/78620/is-the-range-of-a-filterquery-inclusive-up-to-toblock)) to toBlock there were duplicates in every iteration, for example: [0,10000], [10000, 20000]
Would get block 10000 twice.

### Reviewer Checklist
- [x] Code is well-documented
- [] Code deprecates any old functionality before removing it

### DIFF
--- a/chainio/clients/avsregistry/reader.go
+++ b/chainio/clients/avsregistry/reader.go
@@ -21,6 +21,8 @@ import (
 	stakeregistry "github.com/Layr-Labs/eigensdk-go/contracts/bindings/StakeRegistry"
 )
 
+const QueryBlockRange = 10_000
+
 type AvsRegistryReader interface {
 	GetQuorumCount(opts *bind.CallOpts) (uint8, error)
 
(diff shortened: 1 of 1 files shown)


chainio/clients/avsregistry/reader.go
```diff
@@ -384,8 +385,9 @@ func (r *AvsRegistryChainReader) QueryExistingRegisteredOperatorPubKeys(
 	operatorPubkeys := make([]types.OperatorPubkeys, 0)
 
 	// eth_getLogs is limited to a 10,000 range, so we need to iterate over the range
```
samlaf, line 387: can you move this comment above to where you created the `QueryBlockRange = 10_000` const?

samlaf (commented):
LGTM thanks for this. Just one small fix and then we can merge.

chainio/clients/avsregistry/reader.go
taco-paco: Done

samlaf (approved):

samlaf:
ty sir

PR #224, opened 2024-04-29
feat: make bindings, wallet and txManager available

### LINKED ISSUE #220: Create a BuildAll constructor for low-level interactions
currently the clients.BuildAll() function returns high-level chainio clients.
Power users might want direct access to bindings + wallet/txmgr, so we should also make those available

Should prob also write improve the chainio README to show these 2 different ways of interacting with contracts (high-level and low-level).

renlulu:
Fixes #220 .

### Reviewer Checklist
- [ ] Code is well-documented

### DIFF
--- a/chainio/clients/builder.go
+++ b/chainio/clients/builder.go
@@ -35,15 +35,19 @@ type BuildAllConfig struct {
 // for non-instrumented clients that doesn't return metrics/reg, and another instrumented-constructor
 // that returns instrumented clients and the metrics/reg.
 type Clients struct {
-	AvsRegistryChainReader     *avsregistry.AvsRegistryChainReader
-	AvsRegistryChainSubscriber *avsregistry.AvsRegistryChainSubscriber
-	AvsRegistryChainWriter     *avsregistry.AvsRegistryChainWriter
-	ElChainReader              *elcontracts.ELChainReader
-	ElChainWriter              *elcontracts.ELChainWriter
-	EthHttpClient              eth.Client
-	EthWsClient                eth.Client
-	Metrics                    *metrics.EigenMetrics // exposes main avs node spec metrics that need to be incremented by avs code and used to start the metrics server
-	PrometheusRegistry         *prometheus.Registry  // Used if avs teams need to register avs-specific metrics
+	AvsRegistryChainReader      *avsregistry.AvsRegistryChainReader
+	AvsRegistryChainSubscriber  *avsregistry.AvsRegistryChainSubscriber
+	AvsRegistryChainWriter      *avsregistry.AvsRegistryChainWriter
+	ElChainReader               *elcontracts.ELChainReader
+	ElChainWriter               *elcontracts.ELChainWriter
+	EthHttpClient               eth.Client
+	EthWsClient                 eth.Client
+	Wallet                      wallet.Wallet
+	TxManager                   txmgr.TxManager
+	AvsRegistryContractBindings *chainioutils.AvsRegistryContractBindings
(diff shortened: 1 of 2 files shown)


renlulu:
I will rebase after #203 get merged, the diff will be much more less by then.

chainio/clients/builder.go
```diff
@@ -36,15 +37,19 @@ type BuildAllConfig struct {
...
+	AvsRegistryChainSubscriber  *avsregistry.AvsRegistryChainSubscriber
+	AvsRegistryChainWriter      *avsregistry.AvsRegistryChainWriter
+	ElChainReader               *elcontracts.ELChainReader
+	ElChainWriter               *elcontracts.ELChainWriter
+	EthHttpClient               eth.Client
+	EthWsClient                 eth.Client
+	Wallet                      wallet.Wallet
+	TxManager                   txmgr.TxManager
+	AvsRegistryContractBindings *utils.AvsRegistryContractBindings
+	EigenlayerContractBindings  *utils.EigenlayerContractBindings
```
samlaf, line 50: @shrimalmadhur any oppositions to returning everything but the kitchen sink in this constructor?
I was thinking we break it up into smaller builders (for eg one for high-level chainio reader/writer/subscribers, and another for low-level interactions), but I don't see any problem with returning everything in here and letting user take whatever they need?

renlulu: let me know if you prefer break it, i will adjust my PR, thanks!

shrimalmadhur: IMO, 
we should have 3 exported Methods
`BuildELClients` - just returns core EL contract Bindings
`BuildAVSRegistryClients` - just returns AVS resgistry bindings
`BuildAllClients` - returns all
@samlaf how about this and then user can use based on whatever is needed? I don't want someone to force get AVS registry if not needed.

chainio/clients/builder.go
```diff
@@ -180,7 +218,7 @@ func (config *BuildAllConfig) buildElClients(
 		txMgr,
 	)
 
-	return elChainReader, elChainWriter, nil
+	return elChainReader, elChainWriter, avsRegistryContractBindings, elContractBindings, nil
```
shrimalmadhur, line 221: wait why is this retuning `avsRegistryContractBindings` even tho it's not returning any AVS clients?

samlaf: true this shouldn't be here

renlulu: Okie, i am adjusting my PR.

samlaf: lgtm now, thanks

chainio/clients/builder.go
samlaf: https://github.com/Layr-Labs/eigensdk-go/pull/117 this PR does this. I think we can merge that one first then and rebase this one on top?

shrimalmadhur: yes +1 to that.

renlulu: Oh really sorry, i didn't notice that there is a PR already! Just make my PR more clean...

renlulu: I am totally okay you can go with that one first!

samlaf: @renlulu no worries, that was an old one... just merged the other PR. Can you rebase here?

renlulu: GM, just rebased!

samlaf (approved):
This works for now. Don't think it solves what I had in mind for #220 so I will leave that open. But at least we have this for now. Thanks :)

PR #228, opened 2024-05-01
Changed type aliases to type definitions

0xpanicError:
Changed all type aliases in the sdk to type definitions

### What Changed?
here are the types that were changed
``` go
// types/avs.go

type TaskIndex = uint32
type TaskResponseDigest = Bytes32

// types/operator.go

type OperatorAddr = common.Address
type StakeAmount = *big.Int
type OperatorId = Bytes32
type BlockNum = uint32
```

### Reviewer Checklist
- [ ] Code is well-documented


chainio/clients/avsregistry/reader.go
```diff
@@ -235,13 +235,13 @@ func (r *AvsRegistryChainReader) GetOperatorAddrsInQuorumsAtCurrentBlock(
...
 	operatorId types.OperatorId,
-	blockNumber uint32,
+	blockNumber types.BlockNum,
 ) (types.QuorumNums, [][]opstateretriever.OperatorStateRetrieverOperator, error) {
 	quorumBitmap, operatorStakes, err := r.operatorStateRetriever.GetOperatorState0(
 		opts,
 		r.registryCoordinatorAddr,
 		operatorId,
-		blockNumber)
+		uint32(blockNumber))
```
samlaf, line 244: not sure these explicit type casts are the way to go. If we ever change blockNumber to be uint64 for eg then we'll have to go change these everywhere. 

We could define a `UnderlyingType()` on all the types to unwrap them, which we could use here like  `blockNumber.UnderlyingType()`, but not convinced this is nice either.

Need to think about this more and discuss with @shrimalmadhur

0xpanicError: even in the `blsagg_test.go` file there were 15 instances of blockNum defined as `blockNum := uint32(1)` 
The `OperatorStateRetriever` takes in `blockNum` as `uint32` 

should I in the mean time revert the changes made to `blockNum` ?

PR #176, opened 2024-04-04
Integration test scaffold: add anvil chain with deployed contracts to be able to test sdk features like bindings, txmgr, etc

samlaf:
this requires the changes in https://github.com/Layr-Labs/eigenlayer-contracts/pull/495 to be merged first, then create a new commit in eigenlayer-middleware that points to head of contracts, and then we update our eigenlayer-middleware dependency to that.

PR #233, opened 2024-05-06
AVS-215: Add Subgraphs

Sidu28:
Fixes # .

### What Changed?
Adding subgraphs to the services in eigensdk, to replace in memroy data structures.
### Reviewer Checklist
- [ ] Code is well-documented


services/querier.go
samlaf, line 1: Don't think we want a new querier service like this.
We want graph specific versions of the services we have. So for eg a operatorsinfo_subgraph.go in https://github.com/Layr-Labs/eigensdk-go/tree/master/services/operatorsinfo

Sidu28: the idea was this could be used across each of the services

samlaf: What you have here looks like a graphql client. Should either go in a graph/ dir, or maybe in chainio/clients/graph? Also check to see if the existing golang graphql clients like https://github.com/shurcool/graphql and https://github.com/Khan/genqlient don't already do what you need.

Sidu28: oh that does look like what i need, thanks

Sidu28:
@samlaf not sure Im too happy with how complicated the integration seem to be to setup.  Need to import both eigenlayer-middleware and eigenlayer-middleware-offchain as submodules (I re-imported eigenlayer-middleware because for some reason foundry wont allow me to import it from the contracts/lib folder).  I tried to recreate eigenda/inabox's integration test with running exec-commands from the golang test but some of it doesn't seem to be working.  Need to figure out if there's a better way to do all of this, probably try to get a docker container with eveyrthing already setup on it or something like that.  Any ideas?

samlaf:
> I re-imported eigenlayer-middleware because for some reason foundry wont allow me to import it from the contracts/lib folder). 

wdym by import it? eigensdk use eigenlayer-contracts repo via eigenlayer-middleware submodule (eigensdk->middleware->core), so should be the same here? Maybe just the path needs to be added in remappings.txt (or foundry.toml remapping section)?

> I tried to recreate eigenda/inabox's integration test with running exec-commands from the golang test but some of it doesn't seem to be working. Need to figure out if there's a better way to do all of this, probably try to get a docker container with eveyrthing already setup on it or something like that. Any ideas?

Why are you trying to run eigenda's inabox test?
What's the minimal testing setup that you would need here? Just deploy some avs and register an operator? Let's do that instead.
I have https://github.com/Layr-Labs/eigensdk-go/pull/176 that I started to create such an integration framework. Forget what the state of it is though..

services/operatorsinfo/integration_test_deployment/.github/workflows/test.yml
samlaf, line 1: what is this? can github workflows be in subdirs like this? probably copy-pasted from somewhere by mistake?

services/operatorsinfo/integration_test_deployment/src/Counter.sol
samlaf, line 1: delete?

services/operatorsinfo/operatorsinfo_subgraph_test.go
```diff
@@ -0,0 +1,60 @@
...
+			operatorsQueryCalled = true
+			return nil
+		default:
+			return nil
+		}
+	}
+
+	cs := NewOperatorsInfoServiceSubgraph(context.Background(), querier, logger)
+	operatorPubkeys, err := cs.GetOperatorInfo(context.Background(), operatorAddress)
+	assert.True(t, err)
```
samlaf, line 58: assert.NoError instead probably?

services/operatorsinfo/operatorsinfo_subgraph_test.go
```diff
@@ -0,0 +1,60 @@
...
+type mockGraphQLQuerier struct {
+	QueryFn func(ctx context.Context, q any, variables map[string]any) error
+}
+
+func (m mockGraphQLQuerier) Query(ctx context.Context, q any, variables map[string]any) error {
+	return m.QueryFn(ctx, q, variables)
+}
+
+func TestIndexedChainState_GetIndexedOperatorState(t *testing.T) {
+	logger := logging.NewNoopLogger()
```
samlaf, line 23: don't use noop loggers in tests, I did that before but that's a mistake I realized. Because when you run go tests they are silent by default, and only if you add `-v` then they show output. But if we use a noopLogger then even adding `-v` won't print anything.

services/operatorsinfo/operatorsinfo_subgraph_test.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/Layr-Labs/eigensdk-go/logging"
+	"github.com/Layr-Labs/eigensdk-go/types"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/shurcooL/graphql"
+	"github.com/stretchr/testify/assert"
+)
+
+type mockGraphQLQuerier struct {
+	QueryFn func(ctx context.Context, q any, variables map[string]any) error
+}
```
samlaf, line 16: this reads a bit weird. Is it needed to have the QueryFn as a struct field here? Why not define QueryFn as a method? It's only defined once in the test below, so doesn't seem like this kind of abstraction is needed, and its hard to understand why its being defined like this.

services/operatorsinfo/operatorsinfo_subgraph.go
```diff
@@ -0,0 +1,144 @@
...
+		Socket string
+	}
+	G2Point struct {
+		*bn254.G2Affine
+	}
+	G1Point struct {
+		*bn254.G1Affine
+	}
+	GraphQLQuerier interface {
+		Query(ctx context.Context, q any, variables map[string]any) error
```
samlaf, line 52: are the `any`s necessary? If so, can you add comments as to how they are being used and what they are supposed to be. Hard to understand what this is used if we throwaway all type information.

services/operatorsinfo/operatorsinfo_subgraph_integration_test.go
```diff
@@ -0,0 +1,100 @@
...
+	// 	t.Fatal(err)
+	// }
+}
+
+func startAnvilChain() (int, error) {
+	return execCmd("anvil", []string{"--host", "0.0.0.0"}, []string{})
+}
+
+func startGraph() (int, error) {
+	return execCmd("docker", []string{"compose", "up"}, []string{})
```
samlaf, line 42: start anvil and graph using testcontainers like we do everywhere else. This way don't need to rely on other programs being installed on the user's computer, and also version will always be the same so won't have non-deterministic tests.

services/operatorsinfo/operatorsinfo_subgraph_integration_test.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+func TestStartAnvilChain(t *testing.T) {
+	pid, err := startAnvilChain()
+	if err != nil {
+		t.Fatal(err)
+	}
+	fmt.Println("Anvil chain started with PID", pid)
+
+	curr_path, err := os.Getwd()
```
samlaf, line 19: fix lint error by checking err != nil

samlaf (commented):
Generally LGTM. added some syntactic comments

services/operatorsinfo/integration_test_deployment/.github/workflows/test.yml
Sidu28: removed.

services/operatorsinfo/operatorsinfo_subgraph_test.go
Sidu28: err is a boolean value not an actual error.  Changed the variable name

services/operatorsinfo/operatorsinfo_subgraph_test.go
Sidu28: got it.  Anyways not logging anything here, removed it for now

Sidu28: Ok added Zap logger - that should be ok?

services/operatorsinfo/operatorsinfo_subgraph_test.go
samlaf: hmm I see. Still find it weird but chatgpt convinced me its doable.

services/operatorsinfo/operatorsinfo_subgraph_test.go
samlaf: yep

.gitmodules
```diff
@@ -1,3 +1,12 @@
...
 	url = git@github.com:Layr-Labs/eigenlayer-middleware.git
+[submodule "services/operatorsinfo/integration_test_deployment/lib/forge-std"]
+	path = services/operatorsinfo/integration_test_deployment/lib/forge-std
+	url = https://github.com/foundry-rs/forge-std
+[submodule "services/operatorsinfo/integration_test_deployment/lib/eigenlayer-middleware"]
+	path = services/operatorsinfo/integration_test_deployment/lib/eigenlayer-middleware
+	url = https://github.com/Layr-Labs/eigenlayer-middleware
+[submodule "services/operatorsinfo/integration_test_deployment/lib/eigenlayer-middleware-offchain"]
+	path = services/operatorsinfo/integration_test_deployment/lib/eigenlayer-middleware-offchain
+	url = https://github.com/Layr-Labs/eigenlayer-middleware-offchain
```
samlaf, line 12: are these still needed? don't think they are used anymore after the last commit that deleted a bunch of stuff?

samlaf (dismissed):
LGTM. think we can delete the submodules though?

samlaf (approved):

Issue #238, opened 2024-05-09
Separate metrics server into a separate struct


samlaf:
**Is your feature request related to a problem? Please describe.**

Right now the [metrics interface](https://github.com/Layr-Labs/eigensdk-go/blob/7e4891d5099a194ddad62b3afbe83f7e7dcceb3a/metrics/metrics.go#L13) and its implementation combine 2 separate concerns that should be separate:
- one struct to keep the prometheus variables and helper functions to change them
- actual metric server that listens on 9090

Problem comes when multiple services want to register their own metrics.
Each should have its own Metrics struct to keep track of its prom state and change them, but shouldn't also be a metrics server, because then its confusing where the separation of concern is.

**Describe the solution you'd like**

We should have a MetricsServer struct whose only purpose is to take in a registry (or many?) and that has a single StartServer() function to start its server.

**Describe alternatives you've considered**

**Additional context**

PR #239, opened 2024-05-09
Oprs info parallel query

samlaf:
created on top of #238 so should merge that first, which will simplify this PR.

### Reviewer Checklist
- [ ] Code is well-documented

### DIFF
--- a/services/operatorsinfo/operatorsinfo_inmemory.go
+++ b/services/operatorsinfo/operatorsinfo_inmemory.go
@@ -10,6 +10,7 @@ import (
 	"github.com/Layr-Labs/eigensdk-go/crypto/bls"
 	"github.com/Layr-Labs/eigensdk-go/logging"
 	"github.com/Layr-Labs/eigensdk-go/types"
+	"github.com/Layr-Labs/eigensdk-go/utils"
 	"github.com/ethereum/go-ethereum/common"
 )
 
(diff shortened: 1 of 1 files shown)


services/operatorsinfo/operatorsinfo_inmemory.go
```diff
@@ -151,21 +165,35 @@ func (ops *OperatorsInfoServiceInMemory) startServiceInGoroutine(ctx context.Con
...
+		wg.Done()
+	}()
 
-	socketsMap, err := ops.avsRegistryReader.QueryExistingRegisteredOperatorSockets(ctx, nil, nil)
-	if err != nil {
-		ops.logger.Error("Fatal error querying existing registered operator sockets", "err", err, "service", "OperatorPubkeysServiceInMemory")
-		panic(err)
+	wg.Wait()
+	if pubkeysErr != nil {
+		return errors.Join(errors.New("error querying existing registered operators"), pubkeysErr)
```
shrimalmadhur, line 191: we have a wrapErr function in utils now right? should use that

services/operatorsinfo/operatorsinfo_inmemory.go
```diff
@@ -92,7 +102,11 @@ func (ops *OperatorsInfoServiceInMemory) startServiceInGoroutine(ctx context.Con
 			ops.logger.Error("Fatal error opening websocket subscription for new socket registrations", "err", err, "service", "OperatorPubkeysServiceInMemory")
 			panic(err)
 		}
-		ops.queryPastRegisteredOperatorEventsAndFillDb(ctx)
+		err = ops.queryPastRegisteredOperatorEventsAndFillDb(ctx)
+		if err != nil {
+			ops.logger.Error("Fatal error querying past registered operator events and filling db", "err", err, "service", "OperatorPubkeysServiceInMemory")
+			panic(err)
```
shrimalmadhur, line 108: 😭 panic

samlaf: yeah.. didn't want to fix this right now... that will create a bunch more breaking changes if I return error from this function. Will do another day :)

services/operatorsinfo/operatorsinfo_inmemory.go
samlaf: [3f04092](https://github.com/Layr-Labs/eigensdk-go/pull/239/commits/3f040927bc5486d41df9e8f69ef43878c79c82b7)

samlaf: side note: actually don't like utils.wrapErr, wish it was in an errors module, so that we would do errors.wrapErr instead. we could import the sdk errors module as sdkerrors.wrapErr or something

shrimalmadhur (approved):

PR #247, opened 2024-05-17
added derive_operatorid command to egnkey

samlaf:
Not sure if this is the best place to have this but I really need something like this to help me debug wrong operator registration scenarios for tests and local devnets.

Typically generate a bunch of keys from some mnemonic, and want to quickly see their associated operatorId (hash of G1 pubkey associated with that private key). So with this command I'll just be able to pipe the private keys through this to also print their operatorId

### What Changed?

also refactored commands to be in their own packages to prevent var/flag conflicts

### Reviewer Checklist
- [ ] Code is well-documented

### DIFF
--- a/cmd/egnkey/derive_operatorid/derive_operatorid.go
+++ b/cmd/egnkey/derive_operatorid/derive_operatorid.go
@@ -1,0 +1,39 @@
+package derive_operatorid
+
+import (
+	"encoding/hex"
+	"fmt"
+
+	"github.com/Layr-Labs/eigensdk-go/crypto/bls"
+	"github.com/Layr-Labs/eigensdk-go/types"
+	"github.com/urfave/cli/v2"
+)
+
+var (
+	PrivateKey = &cli.StringFlag{
+		Name:     "private-key",
+		Usage:    "(bn254) private key from which to derive operatorId from",
+		Required: true,
+	}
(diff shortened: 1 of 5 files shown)


cmd/egnkey/derive_operatorid/derive_operatorid.go
```diff
@@ -0,0 +1,39 @@
+package derive_operatorid
```
shrimalmadhur, line 1: probably want to avoid this package name 

https://go.dev/blog/package-names

```
Good package names are short and clear. They are lower case, with no under_scores or mixedCaps. They are often simple nouns, such as:

time (provides functionality for measuring and displaying time)
list (implements a doubly linked list)
http (provides HTTP client and server implementations)
The style of names typical of another language might not be idiomatic in a Go program. Here are two examples of names that might be good style in other languages but do not fit well in Go:

computeServiceClient
priority_queue
```

samlaf: changed to `operatorid`: [01fc8ae](https://github.com/Layr-Labs/eigensdk-go/pull/247/commits/01fc8aec9c512f1f4b5dfb98095e2765a21145d2)

shrimalmadhur (approved):

PR #252, opened 2024-05-25
minimal diff to be able to return the task response bytes instead of just the task response digest

afkbyte:
Fixes [AVS-218](https://linear.app/eigenlabs/issue/AVS-218/fix-blsagg-service-to-store-and-return-msgs-not-their-hash)

Update `bls_aggregation` service to return the Task Response bytes 

### What Changed?

- Added the type `TaskResponse` to the avs types and include it in the `BlsAggregationServiceResponse`
- Update `ProcessNewSignature` to take in the `TaskResponse` as a parameter instead of the `TaskReponseDigest` 
- Update `SignedTaskResponseDigest` to return the `TaskResponse` rather than the `TaskResponseDigest`
    - Compute the `TaskResponseDigest` to be signed in the function as the hashFunction of the `TaskResponse` bytes

TODO: update comments

### Reviewer Checklist
- [ ] Code is well-documented

### DIFF
--- a/services/bls_aggregation/blsagg.go
+++ b/services/bls_aggregation/blsagg.go
@@ -274,6 +274,8 @@ func (a *BlsAggregatorService) singleTaskAggregatorGoroutineFunc(
 		select {
 		case signedTaskResponseDigest := <-signedTaskRespsC:
 			a.logger.Debug("Task goroutine received new signed task response digest", "taskIndex", taskIndex, "signedTaskResponseDigest", signedTaskResponseDigest)
+			// Retrieve the TaskResponse from the map
+			taskResponse := a.taskResponseMap[signedTaskResponseDigest.TaskResponseDigest]
 
 			err := a.verifySignature(taskIndex, signedTaskResponseDigest, operatorsAvsStateDict)
 			signedTaskResponseDigest.SignatureVerificationErrorC <- err
--- a/services/bls_aggregation/blsagg.go
+++ b/services/bls_aggregation/blsagg.go
@@ -2,6 +2,7 @@ package blsagg
 
 import (
 	"context"
+	"crypto/sha256"
 	"errors"
 	"fmt"
 	"math/big"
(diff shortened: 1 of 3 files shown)


services/bls_aggregation/blsagg.go
```diff
@@ -134,17 +135,24 @@ type BlsAggregatorService struct {
 	taskChansMutex     sync.RWMutex
 	avsRegistryService avsregistry.AvsRegistryService
 	logger             logging.Logger
+
+	// taskResponseMap is a map of taskResponseDigest to taskResponse
+	taskResponseMap map[types.TaskResponseDigest]types.TaskResponse
```
samlaf, line 140: I'd prefer is this was local to each task's goroutine like the other data structs, just for consistency (also no chance of having concurrency bugs if for eg there's a taskresponsedigest from two different tasks that have the same digest, however unlikely)

types/avs.go
```diff
@@ -8,6 +8,9 @@ import (
 
 type TaskIndex = uint32
 type TaskResponseDigest = Bytes32
+type TaskResponse = interface{}
+
+type TaskResponseHashFunction func(taskResponse TaskResponse) TaskResponseDigest
```
samlaf, line 13: I was originally thinking we'd take/store a SerializedTaskResponse of type `[]byte`, forcing the service user to serialize his task before sending it to us (and he can deserialize when we send him back the response). wydt about doing that instead? any reason why any is needed (I tend to try to stay away from `any` as much as possible).

Also should we use the standard golang Hash interface https://pkg.go.dev/hash#Hash instead of TaskResponseHashFunction? I've personally never used it, but if its already the standard, then we should probably adopt it?

samlaf (commented):
Awesome this looks exactly like what I had in mind.
Left 2 comments based on style, we can discuss.

types/avs.go
samlaf: Thinking about this more. Actually your `any` is nice because then user can just pass whatever struct they use internally.
But then the TaskResponseHashFunction has to both serialize + hash. Actually don't hate this.

services/bls_aggregation/blsagg.go
afkbyte: 2177792fa8f14b23cead9ea6a159cfca2b3c0776

samlaf (approved):

PR #253, opened 2024-05-26
feat: use new NewServeMux instead of Default and ctx for running metrics server

arora-anmol:
### What Changed?
This PR does following:

- introduces new server mux instead of default
- uses context from `Start` to be able to stop server// context handling

### Why this PR?
- `context` arg was currently not being used
- in our e2e tests, we create multiple metric servers (for different services/components/multiple operators). We couldn't register different prometheus registries because of the `DefaultServeMux` coming from `http` package. This change could enable that

### Reviewer Checklist
- [x] Code is well-documented
- [x] Code deprecates any old functionality before removing it

### DIFF
--- a/metrics/eigenmetrics.go
+++ b/metrics/eigenmetrics.go
@@ -8,6 +8,7 @@ import (
 	"github.com/Layr-Labs/eigensdk-go/logging"
 	"github.com/Layr-Labs/eigensdk-go/types"
 	"github.com/Layr-Labs/eigensdk-go/utils"
+	"gorm.io/gorm/logger"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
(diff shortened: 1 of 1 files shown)


arora-anmol:
@samlaf @shrimalmadhur hi guys, made some minor updates to metrics. Let me know if these are helpful, thank you!

samlaf (approved):
Thanks for this!
Also heads up: eventually we want to separate the metrics and server into another separate structs, see https://github.com/Layr-Labs/eigensdk-go/issues/238

arora-anmol:
thanks @samlaf that's a nice idea, happy to take it up if it's okay with you guys

samlaf:
> thanks @samlaf that's a nice idea, happy to take it up if it's okay with you guys

for sure, help is always appreciated :)
cc @dmanc since we had talked about this

PR #256, opened 2024-05-29
add UpdateSocket fn to avs writer client

samlaf:
Fixes # .

### Reviewer Checklist
- [ ] Code is well-documented

### DIFF
--- a/types/operator.go
+++ b/types/operator.go
@@ -71,7 +71,7 @@ func (o Operator) Validate() error {
 // or whatever an avs decides to use
 type Socket string
 
-func (s Socket) ToString() string {
+func (s Socket) String() string {
 	return string(s)
 }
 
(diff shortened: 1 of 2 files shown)


types/operator.go
```diff
@@ -71,6 +71,10 @@ func (o Operator) Validate() error {
 // or whatever an avs decides to use
 type Socket string
 
+func (s Socket) ToString() string {
```
shrimalmadhur, line 74: I think Go idiomatic is `String()` instead of `ToString()` which is more of Java convention.

shrimalmadhur (approved):
Approving this to unblock by check my comment first.

types/operator.go
samlaf: [8d91eb4](https://github.com/Layr-Labs/eigensdk-go/pull/256/commits/8d91eb48eb466d97cad93c4897ed6099ec751dff)

shrimalmadhur (approved):

PR #257, opened 2024-05-30
Update blsagg service hash fn to return error

samlaf:
the hashfn type in blsagg was not returning error, which was forcing implementations to panic.

### What Changed?

Updated to also return error so client can decide what to do on error

### Reviewer Checklist
- [ ] Code is well-documented

### DIFF
--- a/services/bls_aggregation/blsagg.go
+++ b/services/bls_aggregation/blsagg.go
@@ -35,6 +35,9 @@ var (
 	OperatorNotPartOfTaskQuorumErrorFn = func(operatorId types.OperatorId, taskIndex types.TaskIndex) error {
 		return fmt.Errorf("operator %x not part of task %d's quorum", operatorId, taskIndex)
 	}
+	HashFunctionError = func(err error) error {
+		return fmt.Errorf("Failed to hash task response: %w", err)
+	}
 	SignatureVerificationError = func(err error) error {
 		return fmt.Errorf("Failed to verify signature: %w", err)
 	}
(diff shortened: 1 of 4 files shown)


services/bls_aggregation/blsagg.go
```diff
@@ -261,14 +264,19 @@ func (a *BlsAggregatorService) singleTaskAggregatorGoroutineFunc(
...
 			err := a.verifySignature(taskIndex, signedTaskResponseDigest, operatorsAvsStateDict)
 			signedTaskResponseDigest.SignatureVerificationErrorC <- err
 			if err != nil {
 				continue
 			}
+
+			// compute the taskResponseDigest using the hash function
+			taskResponseDigest, err := a.hashFunction(signedTaskResponseDigest.TaskResponse)
+			if err != nil {
+				a.logger.Error("Failed to hash task response, skipping.", "taskIndex", taskIndex, "signedTaskResponseDigest", signedTaskResponseDigest, "err", err)
```
afkbyte, line 277: we already verify that the `taskResponseDigest`  is the result of `hashFunction` in `verifySignature` called before here. Instead of rehashing it here maybe worth matching the error type returned on line 268 and logging that we are skipping due to a failure to hash the task response as a part of that check.

samlaf: Moved log inside verifyFunction: [546d91d](https://github.com/Layr-Labs/eigensdk-go/pull/257/commits/546d91d91e789b9d8d53c302876eca01d344ea75)

We still need to rehash anyways because we need the digest below. As for skipping error check, I don't really like having implicit dependencies on verifySignature's implementation. I'd rather have a bit of duplicate (even if currently useless) code in case someone ever changes the verifySignature implementation or whatnot. I added a comment explaining this.

PR #260, opened 2024-06-06
feat: Add task index in the BlsAggregationServiceResponse when error

fyInALT:
### What Changed?

Add task index in the BlsAggregationServiceResponse when error

When develop aggregator by eigensdk-go, some bls sig task may failed or expired, when this happend, 
the caller will got a err from `blsAggServ.aggregatedResponsesC`.

But this responses will just contain a err by:

```go
	TaskExpiredErrorFn = func(taskIndex types.TaskIndex) error {
		return fmt.Errorf("task %d expired", taskIndex)
	}
```

if the caller should process which task is failed, only can write this code:

```go
	var taskIndex uint32
	if blsAggServiceResp.Err != nil {
		errString := blsAggServiceResp.Err.Error()
		if strings.Contains(errString, "expired") {
			fmt.Sscanf(errString, "task %d expired", &taskIndex)
			t.logger.Debug("expired index", "index", taskIndex)
		}
	} else {
		taskIndex = blsAggServiceResp.TaskIndex
	}

       // process taskIndex ...
```

So we can just put the `taskIndex` into the resp when error, then the caller can not use `Sscanf` which depend by error.

### Reviewer Checklist
- [ ] Code is well-documented

### DIFF
--- a/services/bls_aggregation/blsagg.go
+++ b/services/bls_aggregation/blsagg.go
@@ -235,14 +235,16 @@ func (a *BlsAggregatorService) singleTaskAggregatorGoroutineFunc(
 	operatorsAvsStateDict, err := a.avsRegistryService.GetOperatorsAvsStateAtBlock(context.Background(), quorumNumbers, taskCreatedBlock)
 	if err != nil {
 		a.aggregatedResponsesC <- BlsAggregationServiceResponse{
-			Err: TaskInitializationErrorFn(fmt.Errorf("AggregatorService failed to get operators state from avs registry at blockNum %d: %w", taskCreatedBlock, err), taskIndex),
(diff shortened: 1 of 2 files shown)


samlaf (approved):
Agree we need task index when returning error.
I would prefer we eventually move to custom errors which contain the task index instead of relying on task index being set in the return struct, because that feels error prone (some future developer will add an error somewhere and forget to set the index, and then user code will break).
But I'll accept this for now :)

samlaf:
@shrimalmadhur can you check what the failing test is? Looks like its unrelated to this PR.

shrimalmadhur:
> @shrimalmadhur can you check what the failing test is? Looks like its unrelated to this PR.

It's passed - maybe some intermittent network issue, it is passing in unit-tests job. The only thing which is failing is the wiki thing. @samlaf feel free to merge this.

PR #276, opened 2024-06-20
Simplify register operator avswriter fn

samlaf:
Fixes # .

### What Changed?
Simplify this function which was needlessly complicated. If users need the low level granularity than they can just use the bindings directly is what Im thinking. Most users of the high level writer client probably just want this to happen and not ask too many questions. That's also my case when writing test cases that use this function.

### Reviewer Checklist
- [ ] Code is well-documented

### DIFF
--- a/chainio/clients/avsregistry/writer.go
+++ b/chainio/clients/avsregistry/writer.go
@@ -336,16 +336,16 @@ func (w *AvsRegistryChainWriter) RegisterOperator(
 	var operatorToAvsRegistrationSigSalt [32]byte
 	_, err = rand.Read(operatorToAvsRegistrationSigSalt[:])
 	if err != nil {
-		panic(err)
+		return nil, err
 	}
 
 	curBlockNum, err := w.ethClient.BlockNumber(context.Background())
 	if err != nil {
-		panic(err)
+		return nil, err
 	}
 	curBlock, err := w.ethClient.BlockByNumber(context.Background(), big.NewInt(int64(curBlockNum)))
 	if err != nil {
-		panic(err)
+		return nil, err
 	}
 	sigValidForSeconds := int64(60 * 60) // 1 hour
 	operatorToAvsRegistrationSigExpiry := big.NewInt(int64(curBlock.Time()) + sigValidForSeconds)
--- a/chainio/clients/avsregistry/writer.go
+++ b/chainio/clients/avsregistry/writer.go
@@ -3,6 +3,7 @@ package avsregistry
 import (
 	"context"
 	"crypto/ecdsa"
+	"crypto/rand"
 	"errors"
 	"math/big"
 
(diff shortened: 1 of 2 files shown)


chainio/clients/avsregistry/writer.go
```diff
@@ -32,13 +33,9 @@ type AvsRegistryWriter interface {
 	// socket? RegisterOperatorInQuorumWithAVSRegistryCoordinator is used to register a single operator with the AVS's
 	// registry coordinator. - operatorEcdsaPrivateKey is the operator's ecdsa private key (used to sign a message to
 	// register operator in eigenlayer's delegation manager)
-	//  - operatorToAvsRegistrationSigSalt is a random salt used to prevent replay attacks
-	//  - operatorToAvsRegistrationSigExpiry is the expiry time of the signature
 	RegisterOperatorInQuorumWithAVSRegistryCoordinator(
 		ctx context.Context,
 		operatorEcdsaPrivateKey *ecdsa.PrivateKey,
-		operatorToAvsRegistrationSigSalt [32]byte,
```
shrimalmadhur: is the assumption here is that most users won't want to specify this and are okay using the random salt?

chainio/clients/avsregistry/writer.go
```diff
@@ -32,13 +33,9 @@ type AvsRegistryWriter interface {
 	// socket? RegisterOperatorInQuorumWithAVSRegistryCoordinator is used to register a single operator with the AVS's
 	// registry coordinator. - operatorEcdsaPrivateKey is the operator's ecdsa private key (used to sign a message to
 	// register operator in eigenlayer's delegation manager)
-	//  - operatorToAvsRegistrationSigSalt is a random salt used to prevent replay attacks
-	//  - operatorToAvsRegistrationSigExpiry is the expiry time of the signature
 	RegisterOperatorInQuorumWithAVSRegistryCoordinator(
```
shrimalmadhur, line 42: should be not have a breaking change add another method and keep it deprecated or keep both fn?? I know you would want to keep the function name but think we should avoid breaking ?

chainio/clients/avsregistry/writer.go
samlaf: yep

chainio/clients/avsregistry/writer.go
samlaf: What name would you use?

samlaf: Kept the old one (made it deprecated) and added new function with name RegisterOperator :)
[249e8e8](https://github.com/Layr-Labs/eigensdk-go/pull/276/commits/249e8e87dcff0e25acaad5e34c940b7b568a43d6)

chainio/clients/avsregistry/writer.go
```diff
@@ -236,7 +248,124 @@ func (w *AvsRegistryChainWriter) RegisterOperatorInQuorumWithAVSRegistryCoordina
...
+		PubkeyRegistrationSignature: signedMsg,
+		PubkeyG1:                    G1pubkeyBN254,
+		PubkeyG2:                    G2pubkeyBN254,
+	}
+
+	// generate a random salt and 1 hour expiry for the signature
+	var operatorToAvsRegistrationSigSalt [32]byte
+	_, err = rand.Read(operatorToAvsRegistrationSigSalt[:])
+	if err != nil {
+		panic(err)
```
shrimalmadhur: why Panic? why not return error?

chainio/clients/avsregistry/writer.go
```diff
@@ -236,7 +248,124 @@ func (w *AvsRegistryChainWriter) RegisterOperatorInQuorumWithAVSRegistryCoordina
...
+	}
+
+	curBlockNum, err := w.ethClient.BlockNumber(context.Background())
+	if err != nil {
+		panic(err)
```
shrimalmadhur: why panic?

chainio/clients/avsregistry/writer.go
```diff
@@ -236,7 +248,124 @@ func (w *AvsRegistryChainWriter) RegisterOperatorInQuorumWithAVSRegistryCoordina
...
+	}
+	curBlock, err := w.ethClient.BlockByNumber(context.Background(), big.NewInt(int64(curBlockNum)))
+	if err != nil {
+		panic(err)
```
shrimalmadhur: why panic?

chainio/clients/avsregistry/writer.go
samlaf: good catch. prob copied from somewhere and forgot to change. [276562a](https://github.com/Layr-Labs/eigensdk-go/pull/276/commits/276562a6228b64a05277cdb76f845d1616f0ac6d)

chainio/clients/avsregistry/writer.go
samlaf: [276562a](https://github.com/Layr-Labs/eigensdk-go/pull/276/commits/276562a6228b64a05277cdb76f845d1616f0ac6d)

chainio/clients/avsregistry/writer.go
samlaf: [276562a](https://github.com/Layr-Labs/eigensdk-go/pull/276/commits/276562a6228b64a05277cdb76f845d1616f0ac6d)

shrimalmadhur (approved):

PR #277, opened 2024-06-20
Samlaf avs 559 setup integration test for bls agg

### LINKED ISSUE #261: AVS DevX: Go SDK bls aggregation service doesn't support >1 quorum
**Describe the bug**

bls aggregation needs to add signatures and g2 pubkeys for every quorum that an operator is part of, but it currently only adds them once: https://github.com/Layr-Labs/eigensdk-go/blob/4a204d0e0c9118babf1d74353f72c2b97009d7db/services/bls_aggregation/blsagg.go#L295C31-L295C46

We have unit tests for >1 quorums (eg [here](https://github.com/Layr-Labs/eigensdk-go/blob/4a204d0e0c9118babf1d74353f72c2b97009d7db/services/bls_aggregation/blsagg_test.go#L154)) but those were wrongly implemented, so will need to be fixed. An integration test with the contract would have picked this up.

**To Reproduce**

**Expected behavior**

**Screenshots**

**OS details**

**Additional context**

samlaf:
Fixes # .

### What Changed?

Created one integration test for a single quorum.
We'll be able to use this scaffold to augment this to 2 quorums and fix #261 (AVS-494)

### Reviewer Checklist
- [ ] Code is well-documented


services/bls_aggregation/blsagg_test.go
```diff
@@ -761,6 +773,126 @@ func TestBlsAgg(t *testing.T) {
...
+			NonSignerStakeIndices:        blsAggServiceResp.NonSignerStakeIndices,
+		}
+		avsServiceManager, err := avssm.NewContractMockAvsServiceManager(contractAddrs.ServiceManager, ethHttpClient)
+		require.NoError(t, err)
+		_, _, err = avsServiceManager.CheckSignatures(&bind.CallOpts{}, taskResponseDigest, quorumNumbers.UnderlyingType(), uint32(referenceBlockNumber), nonSignerStakesAndSignature)
+		require.NoError(t, err)
+	})
+
+	t.Run("2 quorums 1 operator", func(t *testing.T) {
+		// TODO: Implement this test
```
shrimalmadhur, line 862: this will come as a separate PR?

services/bls_aggregation/blsagg_test.go
```diff
@@ -761,6 +773,126 @@ func TestBlsAgg(t *testing.T) {
 	})
 }
 
+func TestIntegrationBlsAgg(t *testing.T) {
```
shrimalmadhur, line 775: This test is unreadable. unclear what's happening: probably better to divide/refactor
- setup
- what is actually being tested
- add comments maybe?

services/bls_aggregation/blsagg_test.go
samlaf: yes. there's another issue for it

services/bls_aggregation/blsagg_test.go
samlaf: refactored in [ba79798](https://github.com/Layr-Labs/eigensdk-go/pull/277/commits/ba797983f71fd442c7a2c197cfb60f6ba1a03e6c). Check now, hopefully its easier to read :D

services/bls_aggregation/blsagg_test.go
samlaf: #261

shrimalmadhur (approved):

PR #278, opened 2024-06-21
[ELO-189] refactor el reader/writer to a config based approach

shrimalmadhur:
Fixes # .

### What Changed?
- Add a new config based constructor for EL reader and writer
- Add nil checks so as to avoid nil pointer

### Reviewer Checklist
- [ ] Code is well-documented

### DIFF
--- a/chainio/clients/elcontracts/reader.go
+++ b/chainio/clients/elcontracts/reader.go
@@ -61,7 +61,7 @@ type ELReader interface {
 	) ([32]byte, error)
 }
 
-type ElChainReaderConfig struct {
+type Config struct {
 	DelegationManagerAddress common.Address
 	AvsDirectoryAddress      common.Address
 }
@@ -126,7 +126,7 @@ func BuildELChainReader(
 }
 
 func NewReaderFromConfig(
-	cfg ElChainReaderConfig,
+	cfg Config,
 	ethClient eth.Client,
 	logger logging.Logger,
 ) (*ELChainReader, error) {
--- a/chainio/clients/elcontracts/reader.go
+++ b/chainio/clients/elcontracts/reader.go
@@ -9,7 +9,6 @@ import (
 	gethcommon "github.com/ethereum/go-ethereum/common"
 
 	"github.com/Layr-Labs/eigensdk-go/chainio/clients/eth"
-	chainioutils "github.com/Layr-Labs/eigensdk-go/chainio/utils"
 	"github.com/Layr-Labs/eigensdk-go/logging"
 	"github.com/Layr-Labs/eigensdk-go/types"
 	"github.com/Layr-Labs/eigensdk-go/utils"
@@ -93,13 +92,15 @@ func NewELChainReader(
 	}
 }
 
+// BuildELChainReader creates a new ELChainReader
+// Deprecated: Use BuildFromConfig instead
 func BuildELChainReader(
 	delegationManagerAddr gethcommon.Address,
 	avsDirectoryAddr gethcommon.Address,
 	ethClient eth.Client,
 	logger logging.Logger,
 ) (*ELChainReader, error) {
-	elContractBindings, err := chainioutils.NewEigenlayerContractBindings(
+	elContractBindings, err := NewEigenlayerContractBindings(
 		delegationManagerAddr,
 		avsDirectoryAddr,
 		ethClient,
@@ -118,12 +119,12 @@ func BuildELChainReader(
 	), nil
 }
 
-func BuildELChainReaderFromConfig(
+func BuildFromConfig(
 	cfg types.ElChainReaderConfig,
 	ethClient eth.Client,
 	logger logging.Logger,
 ) (*ELChainReader, error) {
-	elContractBindings, err := chainioutils.NewEigenLayerContractBindingsFromConfig(
+	elContractBindings, err := NewBindingsFromConfig(
 		cfg,
 		ethClient,
 		logger,
--- a/chainio/clients/elcontracts/reader.go
+++ b/chainio/clients/elcontracts/reader.go
@@ -118,11 +118,6 @@ func BuildELChainReader(
 	), nil
 }
 
-type ElChainReaderConfig struct {
-	delegationManagerAddress gethcommon.Address
-	avsDirectoryAddress      gethcommon.Address
-}
-
 func BuildELChainReaderFromConfig(
 	cfg types.ElChainReaderConfig,
 	ethClient eth.Client,
--- a/chainio/clients/elcontracts/reader.go
+++ b/chainio/clients/elcontracts/reader.go
@@ -1,6 +1,7 @@
 package elcontracts
 
 import (
+	"errors"
 	"math/big"
 
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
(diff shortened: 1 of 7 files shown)


types/config.go
```diff
@@ -0,0 +1,8 @@
+package types
+
+import "github.com/ethereum/go-ethereum/common"
+
+type ElChainReaderConfig struct {
+	DelegationManagerAddress common.Address
+	AvsDirectoryAddress      common.Address
+}
```
samlaf: think this should just be in the elchainreader module?

chainio/clients/elcontracts/reader.go
```diff
@@ -129,6 +158,10 @@ func (r *ELChainReader) IsOperatorRegistered(opts *bind.CallOpts, operator types
 }
 
 func (r *ELChainReader) GetOperatorDetails(opts *bind.CallOpts, operator types.Operator) (types.Operator, error) {
+	if r.delegationManager == nil {
+		return types.Operator{}, errors.New("DelegationManager contract not provided")
+	}
```
samlaf, line 170: make this a function if you're going to be repeating everywhere like this

chainio/clients/elcontracts/reader.go
```diff
@@ -116,7 +118,34 @@ func BuildELChainReader(
 	), nil
 }
 
+func BuildELChainReaderFromConfig(
+	cfg types.ElChainReaderConfig,
+	ethClient eth.Client,
+	logger logging.Logger,
+) (*ELChainReader, error) {
```
samlaf: this is literally the same fct as `BuildELChainReader` above, except it takes a cfg instead of the two fcts as arguments. There's literally no difference. I think if you want the cfg to have optional fields you need to take a pointer to the cfg right?

In any case both should behave the same (aka if we go this route, which I'm not super convinced by), then BuildELChainReader should also check whether the addresses are set or not

samlaf (commented):
Not super convinced by this. Honestly I think the better approach unfortunately is to fix the problem upstream, and to go fix the contracts so that they all point to each other and we only need to pass a single contract address, and we can recover all the other contracts from that one contract. Otherwise right now it's a bit confusing that I need to pass DM and AvsDirectory, but if I don't pass the DM, then I also don't get the slasher or SM, but the AvsDirectory is kind of its own little island.

types/config.go
shrimalmadhur: yea I am using this in utils bindings too - putting their creates circular binding I believe.

chainio/clients/elcontracts/reader.go
shrimalmadhur: the problem is - each function has a different return type. so no point of making this as a function?

chainio/clients/elcontracts/reader.go
shrimalmadhur: the difference here is
- In the config nothing is required, you can pass an empty and we won't initialize relevant things.
- Now we can add new contract addresses like rewards coordinator without breaking any method signature.

shrimalmadhur: I am passing the cfg directly into `chainioutils.NewEigenLayerContractBindingsFromConfig` which is checking if contract address is empty or not and initialized based on thaat

shrimalmadhur:
> Not super convinced by this. Honestly I think the better approach unfortunately is to fix the problem upstream, and to go fix the contracts so that they all point to each other and we only need to pass a single contract address, and we can recover all the other contracts from that one contract. Otherwise right now it's a bit confusing that I need to pass DM and AvsDirectory, but if I don't pass the DM, then I also don't get the slasher or SM, but the AvsDirectory is kind of its own little island.

I 100p agree that's the better approach but I feel it's not going to happen soon and we will be stuck with breaking changes anytime we need to add some contracts. we should still go with this even though it's not ideal and make it flexible IMO to unblock. If at any point in future, contracts gets fixed and we have one master registry contract, we could just update the config to have that and deprecate other fields and then just use master contract.

chainio/clients/elcontracts/reader.go
```diff
@@ -73,6 +74,7 @@ type ELChainReader struct {
 // forces EthReader to implement the chainio.Reader interface
 var _ ELReader = (*ELChainReader)(nil)
 
+// TODO(madhur): make this private. All clients should use build functions
```
samlaf: How about we just deprecate it and eventually get rid of this function?

types/config.go
samlaf: let's folow https://go.dev/blog/package-names and stop using utils modules everywhere?
And let's move bindings.go under elcontracts/bindings.go so that they are in the same packge?

chainio/clients/elcontracts/reader.go
```diff
@@ -230,6 +279,10 @@ func (r *ELChainReader) CalculateDelegationApprovalDigestHash(
 func (r *ELChainReader) CalculateOperatorAVSRegistrationDigestHash(
 	opts *bind.CallOpts, operator gethcommon.Address, avs gethcommon.Address, salt [32]byte, expiry *big.Int,
 ) ([32]byte, error) {
+	if r.avsDirectory == nil {
+		return [32]byte{}, errors.New("AVSDirectory contract not provided")
```
samlaf, line 290: let's start implementing custom errors?
Here would be something like
```
type NoContractError struct {
   Contract string
}
```
or maybe even we can have an enum instead of string to make it explicit

chainio/clients/elcontracts/reader.go
samlaf: ok 100% agree. Let's change name to just `reader.BuildFromConfig` to remoe redundancy. We need to change this across the board as well.

chainio/clients/elcontracts/reader.go
shrimalmadhur: yep. sounds good. we can change in separate PR.

types/config.go
shrimalmadhur: Done

chainio/clients/elcontracts/reader.go
shrimalmadhur: Done

chainio/clients/elcontracts/reader.go
shrimalmadhur: discussed - we don't want to - we will have logger.With(component tag)

types/config.go
samlaf: config should be moved from types/config.go to elcontracts/reader.go no?

chainio/clients/elcontracts/reader.go
```diff
@@ -116,7 +119,34 @@ func BuildELChainReader(
 	), nil
 }
 
+func BuildFromConfig(
```
samlaf: Should this just be called `NewFromConfig`? What's the rationale for using Build again?

types/config.go
shrimalmadhur: oh shit. did I not do it. me bad. will do

chainio/clients/elcontracts/reader.go
shrimalmadhur: I don't have a strong pref. not sure why we used build in first place - maybe because it's building the bindings?

types/config.go
shrimalmadhur: moved b9ba1bed85f61239f75c1c5b3d917c30ba4a9553

chainio/clients/elcontracts/reader.go
shrimalmadhur: I have a better names now - b9ba1bed85f61239f75c1c5b3d917c30ba4a9553
```
elcontracts.NewReaderFromConfig
```

I will add similar in future
```
elcontracts.NewWriterFromConfig
avsregistry.NewReaderFromConfig
avsregistry.NewWriterFromConfig
```

wdyt?

samlaf (approved):

Issue #79, opened 2023-11-18
Data race


estensen:
Running this on `master`

```go
go test ./... -race
...
==================
WARNING: DATA RACE
Write at 0x00c00007cde8 by goroutine 17:
  math/big.(*Int).Add()
      /usr/local/Cellar/go/1.21.4/libexec/src/math/big/int.go:147 +0x187
  github.com/Layr-Labs/eigensdk-go/services/bls_aggregation.(*BlsAggregatorService).singleTaskAggregatorGoroutineFunc()
      /Users/estensen/Developer/eigensdk-go/services/bls_aggregation/blsagg.go:261 +0x1c44
  github.com/Layr-Labs/eigensdk-go/services/bls_aggregation.(*BlsAggregatorService).InitializeNewTask.func1()
      /Users/estensen/Developer/eigensdk-go/services/bls_aggregation/blsagg.go:161 +0xf3

Previous read at 0x00c00007cde8 by goroutine 18:
  math/big.(*Int).Mul()
      /usr/local/Cellar/go/1.21.4/libexec/src/math/big/int.go:194 +0x1b5
  github.com/Layr-Labs/eigensdk-go/services/bls_aggregation.checkIfStakeThresholdsMet()
      /Users/estensen/Developer/eigensdk-go/services/bls_aggregation/blsagg.go:370 +0x1ab
  github.com/Layr-Labs/eigensdk-go/services/bls_aggregation.(*BlsAggregatorService).singleTaskAggregatorGoroutineFunc()
      /Users/estensen/Developer/eigensdk-go/services/bls_aggregation/blsagg.go:268 +0x11de
  github.com/Layr-Labs/eigensdk-go/services/bls_aggregation.(*BlsAggregatorService).InitializeNewTask.func1()
      /Users/estensen/Developer/eigensdk-go/services/bls_aggregation/blsagg.go:161 +0xf3

Goroutine 17 (running) created at:
  github.com/Layr-Labs/eigensdk-go/services/bls_aggregation.(*BlsAggregatorService).InitializeNewTask()
      /Users/estensen/Developer/eigensdk-go/services/bls_aggregation/blsagg.go:161 +0x2e6
  github.com/Layr-Labs/eigensdk-go/services/bls_aggregation.TestBlsAgg.func4()
      /Users/estensen/Developer/eigensdk-go/services/bls_aggregation/blsagg_test.go:186 +0x805
  testing.tRunner()
[...]

PR #280, opened 2024-06-24
Test with race detector

estensen:
Would have prevented https://github.com/Layr-Labs/eigensdk-go/issues/79 from being introduced.

Tested locally.

Will make running tests slower, but tests can be made faster by running them in parallel (check `golangci-lint run -E paralleltest`)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ mocks: ## generates mocks
 	go generate ./...
 
 tests: ## runs all tests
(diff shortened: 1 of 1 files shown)


samlaf (dismissed):

samlaf:
My guess is we just need to increase the timeout to 1min here, given that race is making the tests run slowly?

samlaf (approved):

PR #282, opened 2024-06-26
add set claimer for to EL writer

shrimalmadhur:
Fixes # .

### What Changed?
- Add SetClaimerFor function in SDK
- Changing the contracts argument reader and writer struct back to struct instead of interface due to usage restriction for nil checks. We are not using mocks for testing for those so it doesn't make sense to use to interface

### Reviewer Checklist
- [ ] Code is well-documented

### DIFF
--- a/chainio/clients/builder.go
+++ b/chainio/clients/builder.go
@@ -175,11 +175,11 @@ func (config *BuildAllConfig) BuildELClients(
 
 	delegationManagerAddr, err := avsRegistryContractBindings.StakeRegistry.Delegation(&bind.CallOpts{})
 	if err != nil {
-		logger.Fatal("Failed to fetch Slasher contract", "err", err)
+		logger.Fatal("Failed to fetch DelegationManager contract", "err", err)
 	}
 	avsDirectoryAddr, err := avsRegistryContractBindings.ServiceManager.AvsDirectory(&bind.CallOpts{})
 	if err != nil {
-		logger.Fatal("Failed to fetch Slasher contract", "err", err)
+		logger.Fatal("Failed to fetch AVSDirectory contract", "err", err)
 	}
 
 	elContractBindings, err := config.buildEigenLayerContractBindings(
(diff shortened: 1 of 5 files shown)


chainio/clients/builder.go
```diff
@@ -206,6 +206,7 @@ func (config *BuildAllConfig) BuildELClients(
 		elContractBindings.Slasher,
 		elContractBindings.DelegationManager,
 		elContractBindings.StrategyManager,
+		nil,
```
samlaf: add a TODO comment for why this is nil?

chainio/clients/elcontracts/writer.go
```diff
@@ -290,3 +302,30 @@ func (w *ELChainWriter) DepositERC20IntoStrategy(
...
+		return nil, err
+	}
+
+	tx, err := w.rewardsCoordinator.SetClaimerFor(noSendTxOpts, claimer)
+	if err != nil {
+		return nil, err
+	}
+	receipt, err := w.txMgr.Send(ctx, tx)
+	if err != nil {
+		return nil, errors.New("failed to send tx with err: " + err.Error())
```
samlaf: use errors.Join instead

chainio/clients/elcontracts/writer.go
```diff
@@ -290,3 +302,30 @@ func (w *ELChainWriter) DepositERC20IntoStrategy(
...
+
+func (w *ELChainWriter) SetClaimerFor(
+	ctx context.Context,
+	claimer gethcommon.Address,
+) (*gethtypes.Receipt, error) {
+	if w.rewardsCoordinator == nil {
+		return nil, errors.New("RewardsCoordinator contract not provided")
+	}
+
+	w.logger.Infof("setting claimer %s", claimer)
```
samlaf: let's remove all logs in sdk. We don't need this. we give full error context when errors happen anyways.

chainio/clients/elcontracts/writer.go
shrimalmadhur: 599a3176240a861d0d3986348b300a784e596666

chainio/clients/elcontracts/writer.go
shrimalmadhur: 599a3176240a861d0d3986348b300a784e596666

chainio/clients/builder.go
shrimalmadhur: 7110a2a1af0d56c03fe7bbea16e30f52e33b904f

samlaf (approved):

PR #284, opened 2024-06-28
Add gas bumping txmgr

### LINKED ISSUE #153: Update tx-manager-flow diagram
@ian-shim made a bunch of upgrades to our tx-manager->wallet->signer/fireblocks pipeline.

Need to update the diagram in chainio/txmgr/simple-tx-manager-flow.png to reflect this new architecture.

Also need to think through whether getNoSendTxOpts() should still be in txmgr, or should be somewhere else (wallet? maybe in the writer clients directly? unsure about this part..)

samlaf:
Also in passing Fixes #153  .

### What Changed?

Also updated chainio and signer READMEs + some code comments

### Reviewer Checklist
- [ ] Code is well-documented


chainio/txmgr/txmgr.go
```diff
@@ -2,209 +2,35 @@ package txmgr
...
-
-	return types.NewTx(rawTx), nil
+// TxnManager receives transactions from the caller, sends them to the chain, and monitors their status.
+// It also handles the case where a transaction is not mined within a certain time. In this case, it will
+// resend the transaction with a higher gas price. It is assumed that all transactions originate from the
+// same account.
+type AsyncTxManager interface {
+	Start(ctx context.Context)
+	ProcessTransaction(ctx context.Context, req *TxnRequest) error
+	ReceiptChan() chan *ReceiptOrErr
```
samlaf: This I'm not sure what to do with. The eigenda txmgr was implemented using this async interface, whereas our simple txmgr is using a blocking sync interface.

Should we keep them separate like this? Or should we convert one to the other? Def making the simple txmgr into an async version is easier, but is that what we want?

samlaf: resolving because decided to move to sync api

chainio/txmgr/geometric/geometric.go
```diff
@@ -0,0 +1,598 @@
...
+	tx *types.Transaction
+
+	requestedAt time.Time
+	// txAttempts are the transactions that have been attempted to be mined for this request.
+	// If a transaction hasn't been confirmed within the timeout and a replacement transaction is sent,
+	// the original transaction hash will be kept in this slice
+	txAttempts []*transaction
+}
+
+type EthBackend interface {
```
shrimalmadhur: this should be private

chainio/clients/wallet/privatekey_wallet.go
```diff
@@ -3,31 +3,30 @@ package wallet
...
 	"github.com/Layr-Labs/eigensdk-go/signerv2"
 	"github.com/Layr-Labs/eigensdk-go/utils"
-	"github.com/ethereum/go-ethereum/accounts/abi"
-	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
 )
 
-var _ Wallet = (*privateKeyWallet)(nil)
+type EthBackend interface {
```
shrimalmadhur, line 14: This should be private

chainio/clients/wallet/privatekey_wallet.go
```diff
@@ -50,32 +48,17 @@ func (t *privateKeyWallet) SendTransaction(ctx context.Context, tx *types.Transa
 	}
 
 	t.logger.Debug("Sending transaction")
-	opts := &bind.TransactOpts{
```
shrimalmadhur, line 53: <3

chainio/clients/wallet/privatekey_wallet_test.go
```diff
@@ -0,0 +1,70 @@
...
+			To:        &signerAddr,
+			Value:     big.NewInt(1),
+		})
+		ctxWithTimeout, cancel = context.WithTimeout(context.Background(), 5*time.Second)
+		defer cancel()
+		txId, err := skWallet.SendTransaction(ctxWithTimeout, tx)
+		require.NoError(t, err)
+
+		// need to give some time for anvil to process the tx and mine the block
+		// TODO: shall we expose a public WaitForTxReceipt function in the wallet interface, or somewhere else?
```
shrimalmadhur, line 60: the `GetTransactionReceipt` doesn't wait? How would `WaitForTxReceipt ` method diff from that?

chainio/txmgr/geometric/geometric.go
```diff
@@ -0,0 +1,598 @@
...
+	// time to wait for a transaction to be confirmed (mined + confirmationBlocks blocks)
+	// default: 5 * 12 seconds
+	TxnConfirmationTimeout time.Duration
+	// max number of times to retry sending a transaction before failing
+	// this applies to every transaction attempt when a nonce is bumped
+	// default: 3
+	MaxSendTransactionRetry int
+	// time to wait between checking for each transaction receipt
+	// while monitoring transactions to get mined
+	GetTxReceiptTickerDuration time.Duration
```
shrimalmadhur, line 76: is there a default value for this?

chainio/txmgr/geometric/geometric.go
```diff
@@ -0,0 +1,598 @@
...
+func (t *GeometricTxManager) estimateGasFeeCap(ctx context.Context, gasTipCap *big.Int) (*big.Int, error) {
+	header, err := t.ethClient.HeaderByNumber(ctx, nil)
+	if err != nil {
+		return nil, utils.WrapError("failed to get latest header", err)
+	}
+	return new(big.Int).Add(new(big.Int).Mul(header.BaseFee, big.NewInt(2)), gasTipCap), nil
+}
+
+func (t *GeometricTxManager) addGasBuffer(gasLimit uint64) uint64 {
+	return t.params.GasMultiplierPercentage * gasLimit / 100
```
shrimalmadhur: why do we keep GasMultiplier in %age and then divide by 100 instead of just doing it as multiplier. 1.2 instead of 120?

shrimalmadhur: same ques for other multipliers

chainio/txmgr/geometric/geometric.go
```diff
@@ -0,0 +1,598 @@
...
+// Send is used to sign and send a transaction to an evm chain.
+// It does gas estimation and gas price bumping to ensure the transaction gets mined,
+// but it does not do nonce management, so the tx argument must have the correct nonce already set.
+//
+// Send is blocking and safe to call concurrently, so sending multiple txs in parallel is safe.
+func (t *GeometricTxManager) Send(ctx context.Context, tx *types.Transaction) (*types.Receipt, error) {
+	return t.processTransaction(ctx, newTxnRequest(tx))
+}
+
+// ProcessTransaction sends the transaction and runs the monitoring loop which will bump the gasPrice until the tx get
```
shrimalmadhur:
```suggestion
// processTransaction sends the transaction and runs the monitoring loop which will bump the gasPrice until the tx get
```

chainio/txmgr/geometric/geometric.go
```diff
@@ -0,0 +1,598 @@
...
+func (t *GeometricTxManager) processTransaction(ctx context.Context, req *txnRequest) (*types.Receipt, error) {
+	t.logger.Debug("new transaction",
+		"nonce", req.tx.Nonce(), "gasFeeCap", req.tx.GasFeeCap(), "gasTipCap", req.tx.GasTipCap(),
+	)
+	t.metrics.IncrementProcessingTxCount()
+	defer t.metrics.DecrementProcessingTxCount()
+
+	var txn *types.Transaction
+	var txID wallet.TxID
+	var err error
```
shrimalmadhur: you don't need this since line 195 initializes error?

chainio/txmgr/geometric/geometric.go
```diff
@@ -0,0 +1,598 @@
...
+	retryFromFailure := 0
+	from, err := t.wallet.SenderAddress(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get sender address: %w", err)
```
shrimalmadhur: use utils.WrapError everywhere

chainio/txmgr/geometric/geometric.go
```diff
@@ -0,0 +1,598 @@
...
+		gasTipCap, err := t.estimateGasTipCap(ctx)
+		if err != nil {
+			return nil, fmt.Errorf("failed to estimate gas tip cap: %w", err)
+		}
+		txn, err = t.updateGasTipCap(ctx, req.tx, gasTipCap, from)
+		if err != nil {
+			return nil, fmt.Errorf("failed to update gas price: %w", err)
+		}
+		txID, err = t.wallet.SendTransaction(ctx, txn)
+		// TODO: what is this...? does it come from the fireblocks wallet?
```
shrimalmadhur: is this something we are still not sure?

chainio/txmgr/geometric/geometric.go
```diff
@@ -0,0 +1,598 @@
...
+	if err == nil {
+		if receipt.GasUsed > 0 {
+			t.metrics.ObserveGasUsedWei(receipt.GasUsed)
+		}
+		t.metrics.ObserveConfirmationLatencyMs(time.Since(req.requestedAt).Milliseconds())
+	}
+	return receipt, err
+}
+
+// ensureAnyTransactionBroadcasted waits until all given transactions are broadcasted to the network.
```
shrimalmadhur: `ensureAnyTransactionBroadcasted` 
waits until all given transactions are broadcasted to the network

is this comment correct? methods says `any` but description says `all` given txs?

chainio/txmgr/geometric/geometric.go
```diff
@@ -0,0 +1,598 @@
...
+
+func (t *GeometricTxManager) ensureAnyTransactionConfirmed(
+	ctx context.Context,
+	txs []*transaction,
+) (*types.Receipt, error) {
+	queryTicker := time.NewTicker(t.params.GetTxReceiptTickerDuration)
+	defer queryTicker.Stop()
+	var receipt *types.Receipt
+	var err error
+	// transactions that need to be queried. Some transactions will be removed from this map depending on their status.
```
shrimalmadhur, line 297: wdym by `some txs will be removed` and where is it removed??

shrimalmadhur: oh I see the `delete` in line 323

shrimalmadhur: In Java you are never allowed to remove an element while iterating but Golang allows it - interesting - I hope it does not affect iteration lol

chainio/txmgr/geometric/geometric_test.go
```diff
@@ -0,0 +1,373 @@
...
+	backend.startMining()
+	return backend
+}
+
+func (s *fakeEthBackend) WithBaseFeePerGas(baseFeePerGas *big.Int) *fakeEthBackend {
+	s.baseFeePerGas = baseFeePerGas
+	return s
+}
+
+func (s *fakeEthBackend) startMining() {
```
shrimalmadhur, line 271: nice! this is one of the reason I like and hate fakes :P

chainio/txmgr/geometric/metrics.go
```diff
@@ -0,0 +1,142 @@
...
+				Help:       "total transaction confirmation latency summary in milliseconds",
+				Objectives: map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.95: 0.01, 0.99: 0.001},
+			},
+		),
+		gasUsedGwei: promauto.With(reg).NewSummary(
+			prometheus.SummaryOpts{
+				Namespace:  namespace,
+				Subsystem:  subsystem,
+				Name:       "gas_used_total",
+				Help:       "total gas used to submit each transaction onchain",
```
shrimalmadhur, line 60: specify unit wei?

chainio/txmgr/simple.go
```diff
@@ -0,0 +1,201 @@
...
+	// 2*baseFee + gasTipCap makes sure that the tx remains includeable for 6 consecutive 100% full blocks.
+	// see https://www.blocknative.com/blog/eip-1559-fees
+	gasFeeCap := new(big.Int).Add(header.BaseFee.Mul(header.BaseFee, big.NewInt(2)), gasTipCap)
+
+	gasLimit := tx.Gas()
+	// we only estimate if gasLimit is not already set
+	if gasLimit == 0 {
+		from, err := m.wallet.SenderAddress(ctx)
+		if err != nil {
+			return nil, errors.Join(errors.New("send: failed to get sender address"), err)
```
shrimalmadhur, line 174: utils.WrapError everywhere

chainio/clients/wallet/READMD.md
```diff
@@ -0,0 +1,3 @@
+# Wallet
```
ian-shim: typo in the name of the file

chainio/clients/wallet/privatekey_wallet.go
samlaf: [de54bde](https://github.com/Layr-Labs/eigensdk-go/pull/284/commits/de54bdee8272c3c1059bf7601f195589098ae57f)

chainio/clients/wallet/privatekey_wallet_test.go
samlaf: No `GetTransactionReceipt` only calls the eth rpc node endpoint and returns error if txreceipt doesn't exist (tx hasnt been mined) yet. https://github.com/Layr-Labs/eigensdk-go/blob/de54bdee8272c3c1059bf7601f195589098ae57f/chainio/clients/wallet/privatekey_wallet.go#L66

samlaf: There's a [waitForReceipt](https://github.com/Layr-Labs/eigensdk-go/blob/samlaf-avs-588-upstream-tx-manager-to-eigensdk-go/chainio/txmgr/simple.go#L116) private method on the simple txmgr. Wondering if we shouldn't make that interface public and directly on the wallet instead? Or the ethClient interface.... not sure...

chainio/txmgr/geometric/geometric.go
samlaf: [762dd42](https://github.com/Layr-Labs/eigensdk-go/pull/284/commits/762dd42983b0a7e669747469eb234b4ba68f5c7e)

chainio/txmgr/geometric/geometric.go
samlaf: converting to float and back, like `return uint64(t.params.GasMultiplierFloat * float64(gasLimit))` could work, but if the gasLimit is big number (not sure what the cutoff is), there could be floating point rounding errors. chatGPT seems to suggest for float64 numbers, the range 10^16 start having gaps between integers. gasLimit will always be below that (30M/block), so we could potentially use the floating point approach. Do you think that would make the param more easy to understand (setting 1.2 instead of 120)?

chainio/txmgr/geometric/geometric.go
samlaf: [aaa697a](https://github.com/Layr-Labs/eigensdk-go/pull/284/commits/aaa697ac05ebb492f0f1cba584c8eb907bcf71d8)

chainio/txmgr/geometric/geometric.go
samlaf: yeah this stuff I'm still not sure. @ian-shim can you help here. Would like to clean up this code or at least add comments explaining how it works, for future reference.

chainio/txmgr/geometric/geometric.go
samlaf: [f0d08de](https://github.com/Layr-Labs/eigensdk-go/pull/284/commits/f0d08de3414caecfbfcfc21cea1ed15b028a6072)

chainio/txmgr/geometric/geometric.go
samlaf: Good question. @ian-shim I'm also not sure how this function works. It seems to return nil as soon as one function returns an error, which sounds like a bug?
```
		for _, tx := range txs {
			_, err := t.wallet.GetTransactionReceipt(ctx, tx.TxID)
			if err == nil || errors.Is(err, ethereum.NotFound) || errors.Is(err, wallet.ErrReceiptNotYetAvailable) {
				t.metrics.ObserveBroadcastLatencyMs(time.Since(tx.requestedAt).Milliseconds())
				return nil
			}
		}
```
Basically if any one of the txs hasn't been mined (ethereum.NotFound is returned) then we return. But this could also happen if we have a bug and are querying a random (wrong) txHash, so it feels wrong?

This function seems like it should only be called when the wallet is fireblocks? Should we change the code to reflect that?

chainio/txmgr/geometric/geometric.go
samlaf: That's a good question... only found https://groups.google.com/g/golang-nuts/c/rEmaoxi11_A which seems to be say that it's fine... but it's super old. and the official go spec and faq don't seem to mention this anywhere. :\

samlaf: [bb4ae1f](https://github.com/Layr-Labs/eigensdk-go/pull/284/commits/bb4ae1ff3eca55564fbcce796820ec411f109121)
added comment here. maybe some other kind future soul will catch this and know the correct reference

chainio/txmgr/geometric/geometric_test.go
samlaf: what's the reason haha? what's wrong with `startMining`? :P

chainio/txmgr/geometric/metrics.go
samlaf: this is gas, not gasPrice

chainio/txmgr/simple.go
samlaf: [f0d08de](https://github.com/Layr-Labs/eigensdk-go/pull/284/commits/f0d08de3414caecfbfcfc21cea1ed15b028a6072)

chainio/clients/wallet/privatekey_wallet_test.go
shrimalmadhur: but I am not sure if  `waitForReceipt` is a wallet property. since it's always gonna be available via eth node, no matter how we send? hmm unless we consider fireblock state in this too, then it should be on wallet

chainio/txmgr/geometric/geometric_test.go
shrimalmadhur: lol I was just mentioning in general about fakes.
love - because it exactly tells you what you need and gives you thta
hate - you have to write it

chainio/txmgr/geometric/metrics.go
shrimalmadhur: oh whoops. misread. ok

chainio/txmgr/geometric/geometric.go
shrimalmadhur: yea I think `gasMultiplierPercentage` seems overloaded. To me `gasMultiplier` seems simpler to understand.

chainio/clients/wallet/privatekey_wallet_test.go
samlaf: no fireblocks is irrelevant here I think. waiting for a receipt for a txHash is purely polling an eth node for some amount of time waiting for the txReceipt to exist (to have been mined).
I also think it should be some eth node abstraction layer, but we don't have that currently.

shrimalmadhur: ok park it for now. and then we can add it to ethBackend I think or think about it.

chainio/txmgr/geometric/geometric.go
ian-shim: I think this block of code is out of date. Check out https://github.com/Layr-Labs/eigenda/pull/473/files.
The reason why we have to explicitly convert the error is because go http client returns [errors](https://pkg.go.dev/net/http#Client.Do) as `*url.Error` type. Note that `context.DeadlineExceeded` error cannot be matched in this case. To check if the `err` is a timeout error, we have to cover the case where the error is of type `url.Error` and check using `urlErr.Timeout()` method.

chainio/txmgr/geometric/geometric.go
ian-shim: The comment seems wrong. It returns as long as any txn has been broadcasted. 

> This function seems like it should only be called when the wallet is fireblocks?

Correct!

chainio/txmgr/geometric/geometric.go
samlaf: [91d36d1](https://github.com/Layr-Labs/eigensdk-go/pull/284/commits/91d36d1b0df6569da58cb6edd65c89ff296ae644)

chainio/txmgr/geometric/geometric.go
samlaf: Ahh ok I remember now. We had talked about this @shrimalmadhur, I just forgot sorry.
~@ian-shim we basically just converted your explicit urlErr check to using an interface instead, which could work for more types of errors than only the url errors (any error that has a Timeout() method would be caught here).~ EDIT: was wrong.. see below

samlaf: I still have the TODO below which I'm not sure, maybe you have an answer:
```
// TODO: why do we only retry on client or server timeouts? what about other errors?
```

chainio/txmgr/geometric/geometric.go
samlaf: Ok thanks! I fixed the comment [e929bf5](https://github.com/Layr-Labs/eigensdk-go/pull/284/commits/e929bf5feb7d21929b5ede2cb190bc9907d2b83d)
But I'm still a bit confused @ian-shim . waiting for broadcast is only useful for fireblocks, why are we only checking that its a fireblocksWallet [here](https://github.com/Layr-Labs/eigensdk-go/blob/samlaf-avs-588-upstream-tx-manager-to-eigensdk-go/chainio/txmgr/geometric/geometric.go#L375)? Instead of above. Aka why ae we calling ensureAnyTxBroadcasted and doing the whole huge if statement below for other wallets?

But also overall I'm not convinced this function is actually doing something useful. Maybe I'm missing somethingl... but [here](https://github.com/Layr-Labs/eigensdk-go/blob/samlaf-avs-588-upstream-tx-manager-to-eigensdk-go/chainio/txmgr/geometric/geometric.go#L258) we will just return eagerly if any tx has not been mined yet. But not being mined doesn't mean its been broadcasted to the network and is in the mempool right? What we would want is to check if a tx is in the mempool. I think `eth_getTransactionCount` factors in the txs in the mempool, so we could check that the txCount returned is >= the nonce of the tx? Am I missing something here?

chainio/clients/wallet/READMD.md
samlaf: Thanks! [5ff5718](https://github.com/Layr-Labs/eigensdk-go/pull/284/commits/5ff571847abaaea65454314ece0f00f2d526a38b)

chainio/txmgr/geometric/geometric.go
ian-shim: We should only retry on only transient errors but not for other errors. Timeouts are one of them. Probably there are other transient errors, but I wasn't aware of any

chainio/txmgr/geometric/geometric.go
samlaf: @ian-shim after our discussion this is what I came up with: [072d670](https://github.com/Layr-Labs/eigensdk-go/pull/284/commits/072d670dfeb786dbe5cb9ea9bc2e6fe01d20505a)
Can you quickly check that this makes sense?

ian-shim: Thanks, that makes sense!

chainio/txmgr/geometric/geometric.go
samlaf: Ah @ian-shim sorry I was wrong above. Just understood what you meant. Reverted back to your code and added a comment in [9752cca](https://github.com/Layr-Labs/eigensdk-go/pull/284/commits/9752ccad336f17fbcc3b1ea2bbda51caac8f82d7)
This looks fine right?

samlaf: > We should only retry on only transient errors but not for other errors. Timeouts are one of them. Probably there are other transient errors, but I wasn't aware of any

Oh sorry just saw this. I would argue that every error is *possibly* transient. Even a 500 could just be transient, and there's no way to know other than to retry, no?

crypto/ecdsa/utils.go
```diff
@@ -113,3 +114,16 @@ func GetAddressFromKeyStoreFile(keyStoreFile string) (gethcommon.Address, error)
 		return gethcommon.HexToAddress(address), nil
 	}
 }
+
+func KeyAndAddressFromHexKey(hexkey string) (*ecdsa.PrivateKey, common.Address, error) {
+	if len(hexkey) > 2 && hexkey[:2] == "0x" {
```
ian-shim:
```
hexkey = strings.TrimPrefix(hexkey, "0x")
hexkey = strings.TrimPrefix(hexkey, "0X")
```

chainio/txmgr/metrics.go
```diff
@@ -0,0 +1,108 @@
...
+
+import (
+	"github.com/Layr-Labs/eigensdk-go/logging"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+const namespace = "txmgr"
+
+type Metrics struct {
```
ian-shim: How is this different from the one under geometric?

chainio/txmgr/geometric/geometric.go
```diff
@@ -0,0 +1,609 @@
...
+}
+
+func NewGeometricTxnManager(
+	ethClient ethBackend,
+	wallet wallet.Wallet,
+	logger logging.Logger,
+	metrics Metrics,
+	params GeometricTxnManagerParams,
+) *GeometricTxManager {
+	fillParamsWithDefaultValues(&params)
```
ian-shim: Not sure if enforcing the default values here makes the most sense. 
How about providing the default params as exported value so callers can use it, and just validating these params in the constructor (returning error if any params missing)?

chainio/txmgr/geometric/geometric.go
```diff
@@ -0,0 +1,609 @@
...
+		gasTipCap, err := t.estimateGasTipCap(ctx)
+		if err != nil {
+			return nil, utils.WrapError("failed to estimate gas tip cap", err)
+		}
+		txn, err = t.updateGasTipCap(ctx, req.tx, gasTipCap, from)
+		if err != nil {
+			return nil, utils.WrapError("failed to update gas price", err)
+		}
+		txID, err = t.wallet.SendTransaction(ctx, txn)
+		// fireblocks wallet uses go's net.Http client which returns a url.Error on timeouts
```
ian-shim: geth also uses http client

chainio/txmgr/geometric/geometric.go
```diff
@@ -0,0 +1,609 @@
...
+			continue
+		} else if err != nil {
+			return nil, utils.WrapError(fmt.Errorf("failed to send txn %s", txn.Hash().Hex()), err)
+		} else {
+			t.logger.Debug("successfully sent txn", "txID", txID, "txHash", txn.Hash().Hex())
+			break
+		}
+	}
+
+	// is this case even possible? we return on errors above
```
ian-shim: Yes if all retries failed

chainio/txmgr/geometric/geometric.go
```diff
@@ -0,0 +1,609 @@
...
+// but it does not do nonce management, so the tx argument must have the correct nonce already set.
+//
+// Send is blocking and safe to call concurrently, so sending multiple txs in parallel is safe.
+func (t *GeometricTxManager) Send(ctx context.Context, tx *types.Transaction) (*types.Receipt, error) {
+	return t.processTransaction(ctx, newTxnRequest(tx))
+}
+
+// processTransaction sends the transaction and runs the monitoring loop which will bump the gasPrice until the tx get
+// included.
+func (t *GeometricTxManager) processTransaction(ctx context.Context, req *txnRequest) (*types.Receipt, error) {
```
ian-shim, line 181: Is this method thread safe? 
Could a transaction with higher nonce end up published before one with lower nonce?

crypto/ecdsa/utils.go
samlaf: [6ae0800](https://github.com/Layr-Labs/eigensdk-go/pull/284/commits/6ae08003eee99b279ede1ec44ecb2ea2b8961a74)

chainio/txmgr/metrics.go
samlaf: Good catch! Think I was trying to make it generic initially, but then realized the metrics are specific to geometric. That file was outdated, deleted it. [0e98f3a](https://github.com/Layr-Labs/eigensdk-go/pull/284/commits/0e98f3a4632fd84ba52c748ac167702561503c39)

chainio/txmgr/geometric/geometric.go
samlaf: They are not enforced! They are only set when the value is unset (has the zero value, which doesnt make sense for all params). Changed the name to reflect that: [09d3e23](https://github.com/Layr-Labs/eigensdk-go/pull/284/commits/09d3e23a483cbe3e5d5081b09a2113c71f20c4ca)

chainio/txmgr/geometric/geometric.go
samlaf: Oh snap you're right... https://github.com/ethereum/go-ethereum/blob/83e70aa3d00bbee9713b31768a30a7741eee8945/rpc/http.go#L231
Huh why is error handling so ugly lol? I remember reading https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully and thinking I had it figured out by using error interfaces everywhere, but clearly I haven't haha.

So hmmm... shouldn't every package just wrap the underlying package's errors and redefine their own? Basically abstracting errors so that their clients don't need to deal with the error dependencies of 5 package dependencies down?

samlaf: Also I'm confused when is context.DeadlineExceeded returned then? If they all use http.Do, wouldn't they all return url.Error with a Timeout() method instead of context.DeadlineExceeded?

samlaf: changed the comment to [0219820](https://github.com/Layr-Labs/eigensdk-go/pull/284/commits/02198207e7af8b60f7abf044c5657aeffa74278c)
Let me know if this is better (at least hopefully not wrong).

chainio/txmgr/geometric/geometric.go
samlaf: [e8fa6a0](https://github.com/Layr-Labs/eigensdk-go/pull/284/commits/e8fa6a00d2340ccb48c6838049aa3f01a21aafcc)

chainio/txmgr/geometric/geometric.go
ian-shim: Invalid requests should not be retried for sure. 
Harder to tell if a particular 500 error is transient. What we do for EigenDA is the eth client itself retries a request with a different provider if it receives a 5xx error.

chainio/txmgr/geometric/geometric.go
ian-shim: What I meant was, the constructor sets the params under the hood (if some values are missing) and the caller may not realize that. Was thinking it would be better to be explicit

chainio/txmgr/geometric/geometric.go
ian-shim: > If they all use http.Do, wouldn't they all return url.Error with a Timeout() method instead of context.DeadlineExceeded?

We don't know if they all return `url.Error` 😅

chainio/txmgr/geometric/geometric.go
samlaf: I don't think so. It's very typical to have a bunch of optional parameters with defaults. aws sdk has it for eg. One nice way to set them is with the functional options pattern: https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis.
But if there's a lot of parameters like here, it's just very annoying to have to set all of them explicitly, especially when you don't know what to set. And having to write 10 default params is just annoying I feel.

chainio/txmgr/geometric/geometric.go
samlaf: Yes it is. Added a test to show this: [549b045](https://github.com/Layr-Labs/eigensdk-go/pull/284/commits/549b045cde1f18ddb3f2f62a3a5302ae48664ab2)
Also added comment to explain how to send txs concurrently: [e6ab890](https://github.com/Layr-Labs/eigensdk-go/pull/284/commits/e6ab89049cbe3c08b4bbca7869afeae0c1e8ca15)

samlaf: Oh btw also had to add https://github.com/Layr-Labs/eigensdk-go/pull/284/commits/3a12fcc160284c6ad2d62dbdb1bdd811dccfcab2#diff-f3751b0e1d13e5e4c58c0ae0d936505427becc99a1c52514628bdaf32df6d370R488 to slow down the gas bumping loop, otherwise txs were getting bumped like crazy if a txs with the wrong (too-high) nonce was sent. Is that normal? Was there a better way to deal with this?

chainio/txmgr/geometric/geometric.go
```diff
@@ -0,0 +1,623 @@
...
+				} else {
+					t.logger.Debug("failed to get chain tip while waiting for transaction to mine", "err", err)
+				}
+			}
+
+			// TODO(samlaf): how to maintain these better? How do we know which errors to use and where they are
+			// returned from?
+			if errors.Is(err, ethereum.NotFound) || errors.Is(err, wallet.ErrReceiptNotYetAvailable) {
+				// t.logger.Debug("Transaction not yet mined", "nonce", tx.Nonce(), "txID", txID, "txHash",
+				// tx.Hash().Hex(), "err", err)
```
ian-shim: What do we want to do here? We can remove this block if nothing is being done

chainio/txmgr/geometric/geometric.go
```diff
@@ -0,0 +1,623 @@
...
+			} else if errors.Is(err, wallet.ErrTransactionFailed) {
+				t.logger.Debug("Transaction failed", "txID", txID, "txHash", tx.Hash().Hex(), "err", err)
+				// Remove the transaction from the list of transactions to query.
+				// go seemingly allows deleting from a map while iterating over it:
+				// https://groups.google.com/g/golang-nuts/c/rEmaoxi11_A
+				// although the official spec and faq don't seem to mention this anywhere...
```
ian-shim, line 338: https://go.dev/doc/effective_go#for

ian-shim (dismissed):
lgtm, I'd integrate this with some service (avssync?) and soak test it as there might be more issues when dealing with the real chain in different gas market

chainio/txmgr/geometric/geometric.go
samlaf: This doesn't seem to mention that deleting items while iterating is safe?

chainio/txmgr/geometric/geometric.go
samlaf: oops commented while debugging and committed by mistake. Uncommented in [63832fc](https://github.com/Layr-Labs/eigensdk-go/pull/284/commits/63832fc803ebdbc196645d25ce56f33ab776e35b)

ian-shim (approved):

PR #286, opened 2024-07-02
move builders to it's own package

shrimalmadhur:
Fixes # .

### Reviewer Checklist
- [ ] Code is well-documented


chainio/clients/elcontracts/builder.go
```diff
@@ -0,0 +1,49 @@
...
+	elContractBindings, err := NewBindingsFromConfig(
+		config,
+		ethHttpClient,
+		logger,
+	)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+
+	elChainReader := NewELChainReader(
```
samlaf: didn't we have new names for these without EL? `elcontracts.NewChainReader`?

shrimalmadhur: updated

shrimalmadhur: lol sorry for no commit info :P

chainio/clients/avsregistry/builder.go
```diff
@@ -0,0 +1,72 @@
...
+
+	chainReader := NewChainReader(
+		avsBindings.RegistryCoordinatorAddr,
+		avsBindings.BlsApkRegistryAddr,
+		avsBindings.RegistryCoordinator,
+		avsBindings.OperatorStateRetriever,
+		avsBindings.StakeRegistry,
+		logger,
+		client,
+	)
```
samlaf, line 35: you have a chainReader here and creating another elChainReader below. Is that normal?

services/bls_aggregation/blsagg.go
```diff
@@ -229,13 +229,13 @@ func (a *BlsAggregatorService) singleTaskAggregatorGoroutineFunc(
...
 ) {
 	defer a.closeTaskGoroutine(taskIndex)
-
 	quorumThresholdPercentagesMap := make(map[types.QuorumNum]types.QuorumThresholdPercentage)
 	for i, quorumNumber := range quorumNumbers {
 		quorumThresholdPercentagesMap[quorumNumber] = quorumThresholdPercentages[i]
 	}
 	operatorsAvsStateDict, err := a.avsRegistryService.GetOperatorsAvsStateAtBlock(context.Background(), quorumNumbers, taskCreatedBlock)
 	if err != nil {
+		a.logger.Error("Task goroutine failed to get operators state from avs registry", "taskIndex", taskIndex, "err", err)
```
shrimalmadhur, line 238: @samlaf I added the error logs here for now since this is useful to add for in case anything fails

chainio/clients/avsregistry/builder.go
```diff
@@ -0,0 +1,72 @@
...
+		config,
+		client,
+		logger,
+	)
+
+	if err != nil {
+		return nil, nil, nil, nil, err
+	}
+
+	chainReader := NewChainReader(
```
samlaf, line 27: why not FromConfig like for subscriber?

chainio/clients/avsregistry/builder.go
```diff
@@ -0,0 +1,72 @@
...
+			AvsDirectoryAddress:      avsBindings.AvsDirectoryAddr,
+		},
+		client,
+		logger,
+	)
+	if err != nil {
+		return nil, nil, nil, nil, err
+	}
+
+	chainWriter := NewChainWriter(
```
samlaf, line 59: why not FromConfig to be consistent?

chainio/clients/avsregistry/builder.go
```diff
@@ -0,0 +1,72 @@
...
+
+	// This is ugly, but we need elReader to be able to create the AVS writer
+	elChainReader, err := elcontracts.NewReaderFromConfig(
+		elcontracts.Config{
+			DelegationManagerAddress: avsBindings.DelegationManagerAddr,
+			AvsDirectoryAddress:      avsBindings.AvsDirectoryAddr,
+		},
+		client,
+		logger,
+	)
```
samlaf, line 54: do we *really* need this to be created outside and passed via dependency injection? Or could we just build it inside the NewChainWriter constructor?

chainio/clients/builder.go
```diff
@@ -87,27 +87,35 @@ func BuildAll(
 		return nil, utils.WrapError("Failed to create transaction sender", err)
 	}
 	txMgr := txmgr.NewSimpleTxManager(pkWallet, ethHttpClient, logger, addr)
-	// creating EL clients: Reader, Writer and EigenLayer Contract Bindings
-	elChainReader, elChainWriter, elContractBindings, err := config.BuildELClients(
+
+	// creating AVS clients: Reader and Writer
+	avsRegistryChainReader, avsRegistryChainSubscriber, avsRegistryChainWriter, avsRegistryContractBindings, err := avsregistry.BuildClients(
```
samlaf, line 92: should BuildClients have a FromConfig suffix?

samlaf (commented):
LGTM. Just some small nits wrt being consistent with the FromConfig or not. Think we should use the same convention everywhere?

chainio/clients/avsregistry/builder.go
shrimalmadhur: we can but we need bindings for both reader and writer so I am just creating binding once using `NewBindingsFromConfig` and re-using that. otherwise the bindings gets created twice.

chainio/clients/avsregistry/builder.go
shrimalmadhur: hmm good ques. idk. need to think more

chainio/clients/avsregistry/builder.go
shrimalmadhur: we can but we need bindings for both reader and writer so I am just creating binding once using `NewBindingsFromConfig` and re-using that. otherwise the bindings gets created twice.

chainio/clients/builder.go
shrimalmadhur: would there by any other way we would want users to build? I guess we don't know yet. just want to keep smaller names lol

samlaf (approved):

PR #292, opened 2024-07-05
move some warn to debug

shrimalmadhur:
Fixes # .

### What Changed?
- Making that warn to debug to not alarm users that something is wrong. If they call the methods of contract which they did not provide, they will get error anyway. 

### Reviewer Checklist
- [ ] Code is well-documented


samlaf (approved):
This is fine. I don't feel 100% convinced this is the right move here though, and expecting this hidden debug warning to result in some user pain at some point haha, but let's see. If we get issues we can change later or figure out a strategy.

PR #296, opened 2024-07-12
fix slog logger constructor nil pointer bug

samlaf:
Fixes # .

### Reviewer Checklist
- [ ] Code is well-documented

### DIFF
--- a/logging/slog_logger.go
+++ b/logging/slog_logger.go
@@ -42,17 +42,32 @@ type SLoggerOptions struct {
 	NoColor bool
 }
 
+// default SLogger options are used when no options are provided
+// they are the development options (debug logs with source)
+var defaultTintOptions = tint.Options{
+	AddSource:   true,
+	Level:       slog.LevelDebug,
+	ReplaceAttr: nil,
+	TimeFormat:  time.StampMilli,
+	NoColor:     false,
+}
+
 // NewSlogTextLogger creates a new SLogger with a text handler
 // Default behavior is colored log outputs. To disable colors, set opts.NoColor to true.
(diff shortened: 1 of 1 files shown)


logging/slog_logger.go
```diff
@@ -42,17 +42,32 @@ type SLoggerOptions struct {
...
-		Level:       opts.Level,
-		ReplaceAttr: opts.ReplaceAttr,
-		TimeFormat:  opts.TimeFormat,
-		NoColor:     opts.NoColor,
+	var tintOptions tint.Options
+	if opts == nil {
+		tintOptions = defaultTintOptions
+	} else {
+		tintOptions = tint.Options{
+			AddSource:   opts.AddSource,
```
shrimalmadhur, line 63: What is opts and not nil and only one of the field is provided. Shouldn't other fields resolve to default options ?

samlaf: Hmm indeed this is a tough one.
If fields are not provided go will still set them to their zero value, and then there's no way for us to know whether they weren't passed, or were explicitly set to zero by the user.

For eg there's no way to differentiate a user having passed `tint.Options{AddSource: false}` or `tint.Options{}` because go will anyways make AddSource the boolean zero value (false) when not passed.

shrimalmadhur: yea so the resolver way of things we have seen - where it would resolve to default value (in this case, it's `false`) which is fine. I don't think we need to worry about that if it was false by default or user provided it since the end result will be `false`. We only care about that if user provide `true` then it should be set of `true`

shrimalmadhur (approved):

PR #297, opened 2024-07-14
refactored elcontracts and avsregistry builders to use wire dependency injection

samlaf:
See https://github.com/google/wire/tree/main
Overall thinking this could greatly simplify some of our code.
So far I've only refactored basic code to use this, so it feels useless.
But it will start paying off once we use it to update the chainio/clients/builder.go. That way we'll be able to add a bunch of different constructors, all automatically generated, for different use cases:
- with privatekey signer vs remote signer
- with or without metrics
- with mocks
- etc

### Reviewer Checklist
- [ ] Code is well-documented


Makefile
```diff
@@ -13,8 +13,9 @@ help:
 bindings: ## generates contract bindings
 	cd contracts && rm -rf bindings/* && ./generate-bindings.sh 
 
-mocks: ## generates mocks
+generate: ## generates mocks and wire files
```
shrimalmadhur, line 16: why do we need to couple these 2 generation? any particular reason or just to combine generation together?

samlaf: Is there a way to separate go generated stuff? Right now if we run go generate it just looks for all the go:generate comments in the repo and executed all of them. Maybe there’s a way to tag the generate commands with subtask names and get go generate to filter on those?

samlaf:
Closing as no longer needed. Feels like not enough leverage for the kind of work/refactor needed to push this across the codebase right now.

PR #301, opened 2024-07-18
refactor: remove interfaces of el reader writer

shrimalmadhur:
Fixes # .

### What Changed?
- Remove unnecessary interface from EL contracts. Clients who use it can create their own interface and use methods as they need. 

### Reviewer Checklist
- [ ] Code is well-documented


metrics/collectors/economic/economic.go
```diff
@@ -61,8 +60,13 @@ type Collector struct {
 
 var _ prometheus.Collector = (*Collector)(nil)
 
+//go:generate mockgen -destination=./mocks/elReader.go -package=mocks github.com/Layr-Labs/eigensdk-go/metrics/collectors/economic ELReader
```
samlaf: why are you generating this? Is this mock used in the tests somewhere? Why not create a fake instead if so?

shrimalmadhur: yes check `economic_test`. The problem with fake again is, I cannot return what I want. I can only return single return value. what if my two tests expect different return value? mock let me return a value I want. check this https://github.com/Layr-Labs/eigensdk-go/pull/301/files#diff-3191b73ec7b1e1c6b0adf55013bf00c38e8ace491aa562328a1388e7ebffd002R29-R30

shrimalmadhur: 8bbba4610bbaf3ddabad884739b743aa842ef4ec

samlaf (approved):

PR #302, opened 2024-07-18
chore: mock to fake

shrimalmadhur:
Fixes # .

### Reviewer Checklist
- [ ] Code is well-documented


services/operatorsinfo/operatorsinfo_inmemory_test.go
```diff
@@ -2,25 +2,115 @@ package operatorsinfo
...
 	"github.com/Layr-Labs/eigensdk-go/types"
 	"github.com/ethereum/go-ethereum/common"
 	gethtypes "github.com/ethereum/go-ethereum/core/types"
-	"go.uber.org/mock/gomock"
-
-	apkregistrybindings "github.com/Layr-Labs/eigensdk-go/contracts/bindings/BLSApkRegistry"
-	regcoordbindings "github.com/Layr-Labs/eigensdk-go/contracts/bindings/RegistryCoordinator"
 )
 
+type fakeAVSRegistryReader struct {
```
shrimalmadhur: @samlaf @eigen-vi my beef with at least these fake is, I am doing all these work for mimicking the same behavior what mocks would do. I am actually writing these fake to return what I what which is what mock does. Input to any function is not at all used (maybe that's what is wrong?) - all processing happens in constructor and then rest is just a read.

shrimalmadhur: sycned with @eigen-vi - convinced on fakes. I feel this stuff right now is not complex so it's annoying but eventually I think fakes would pay off.

chainio/clients/avsregistry/writer.go
```diff
@@ -239,6 +176,18 @@ func NewWriterFromConfig(
 	), nil
 }
 
+// RegisterOperatorInQuorumWithAVSRegistryCoordinator
+// TODO(samlaf): an operator that is already registered in a quorum can register with another quorum without passing
+// signatures perhaps we should add another sdk function for this purpose, that just takes in a quorumNumber and
+// socket? RegisterOperatorInQuorumWithAVSRegistryCoordinator is used to register a single operator with the AVS's
```
samlaf, line 182: why is this green (new)? did you move it from somewhere? I feel like I had written that at some point haha

internal/fakes/avs_regsitry.go
samlaf, line 1: file typo: avs_regsitry -> avs_registry

internal/fakes/avs_regsitry.go
```diff
@@ -0,0 +1,100 @@
+package fakes
+
+import (
+	"context"
+	"math/big"
+
+	apkregistrybindings "github.com/Layr-Labs/eigensdk-go/contracts/bindings/BLSApkRegistry"
+	opstateretrievar "github.com/Layr-Labs/eigensdk-go/contracts/bindings/OperatorStateRetriever"
```
samlaf: typo: retrievar -> retriever

samlaf (dismissed):

chainio/clients/avsregistry/writer.go
shrimalmadhur: because we removed the interface definition from here - to keep the comments I moved it on top of struct implementation.

internal/fakes/avs_regsitry.go
shrimalmadhur: haha - copy pasta mistake

internal/fakes/avs_regsitry.go
shrimalmadhur: 000082458e50457b6388181918262887c5fd3941

internal/fakes/avs_regsitry.go
shrimalmadhur: 000082458e50457b6388181918262887c5fd3941

samlaf (approved):

PR #305, opened 2024-07-23
update: eigensdk bindings for operator set

shrimalmadhur:
Fixes # .

### What Changed?
- Op set bindings

### Reviewer Checklist
- [ ] Code is well-documented


samlaf:
@shrimalmadhur is "opset" the terminology used in our contracts and/or docs? If not, let's use the full name so as not to confuse anyone else reading these.

samlaf (approved):
LGTM. The weird abigen errors we should eventually investigate but let's go with the sleeps for now

PR #311, opened 2024-07-29
feat: add eigenpod bindings

shrimalmadhur:
Fixes # .

### Reviewer Checklist
- [x] Code is well-documented

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,10 @@ help:
 	@grep -E '^[a-zA-Z0-9_-]+:.*?## .*$$' $(MAKEFILE_LIST) | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
 
 bindings: ## generates contract bindings
-	cd contracts && rm -rf bindings/* && ./generate-bindings.sh 
+	cd contracts && rm -rf bindings/* && ./generate-bindings.sh
+
+eigenpod-bindings: ## generates contract bindings for eigenpod
+	cd chainio/clients/eigenpod && ./generate.sh
 
 mocks: ## generates mocks
 	go install go.uber.org/mock/mockgen@v0.4.0
--- a/chainio/clients/eigenpod/generate.sh
+++ b/chainio/clients/eigenpod/generate.sh
@@ -1,6 +1,12 @@
 #!/bin/bash
 
-TMP_DIR=tmp
+# This script generates Go bindings for the EigenPods contracts
+# The reason this script is separate from the bindings generation in the ~/contracts directory
+# is because they are in a separate branch in eigenlayer-contracts and to unblock the development
+# of the EigenPods client, we need to generate the bindings for the contracts in the feat/partial-withdrawal-batching branch.
+# Once eigenlayer-contracts repo is stable and features are on single branch, we can move the bindings back to normal process
+
+TMP_DIR=$(mktemp -d)
 BINDINGS_DIR=bindings
 
 cleanup_bindings_dir() {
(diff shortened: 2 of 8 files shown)


Makefile
```diff
@@ -11,7 +11,10 @@ help:
 	@grep -E '^[a-zA-Z0-9_-]+:.*?## .*$$' $(MAKEFILE_LIST) | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
 
 bindings: ## generates contract bindings
-	cd contracts && rm -rf bindings/* && ./generate-bindings.sh 
+	cd contracts && rm -rf bindings/* && ./generate-bindings.sh
```
samlaf, line 14: think bindings should also generate the eigenpod-bindings, to make sure we don't forget?

chainio/clients/eigenpod/reader.go
```diff
@@ -0,0 +1,74 @@
...
+	logger    logging.Logger
+	ethClient eth.HttpBackend
+	*bindings.IEigenPodCaller
+}
+
+type ManagerChainReader struct {
+	logger    logging.Logger
+	ethClient eth.HttpBackend
+	*bindings.IEigenPodManagerCaller
+}
```
samlaf: why not merge them in a single struct like we do everything else?
If there's a good reason then add as comment above the structs

chainio/clients/eigenpod/reader.go
```diff
@@ -0,0 +1,74 @@
...
+
+func NewReader(
+	address common.Address,
+	ethClient eth.HttpBackend,
+	logger logging.Logger,
+) (*ChainReader, error) {
+	pod, err := NewContractCallerBindings(address, ethClient)
+	if err != nil {
+		return nil, err
```
samlaf: wrap errors with context everywhere

chainio/clients/eigenpod/reader.go
```diff
@@ -0,0 +1,74 @@
...
+
+	return &ManagerChainReader{
+		logger:                 logger,
+		ethClient:              ethClient,
+		IEigenPodManagerCaller: manager,
+	}
+}
+
+func NewReader(
+	address common.Address,
```
samlaf:
```suggestion
	eigenpodAddress common.Address,
```
just having address is error prone I feel. Or at least add a comment explaining whta address this is

chainio/clients/eigenpod/generate.sh
```diff
@@ -0,0 +1,57 @@
...
+
+cleanup_bindings_dir() {
+  echo "Cleaning up the bindings directory"
+  rm -rf ${BINDINGS_DIR}
+}
+
+create_tmp_dir() {
+  echo "Creating a temporary directory"
+  mkdir -p ${TMP_DIR}
+}
```
samlaf: bash has a function for this which is pretty cool, also no risk of poisoning your local directory.
```
local tmpdir=$(mktemp -t mydir)
```
this creates a directory in `/tmp`

chainio/clients/eigenpod/generate.sh
samlaf, line 1: I would add a comment somewhere, whether in a chainio/clients/eigenpod/README, or in this file, or in the makefile, explaining why we generate bindings separately, and we don't use the submodule that we use for everything else, etc.

chainio/clients/eigenpod/reader.go
shrimalmadhur: yea there is. most people who want to manage an eigenpod only need the first one. Eigenpod manager is mostly for creating a new eigenpod. I will add that comment.

Makefile
shrimalmadhur: it does generate https://github.com/Layr-Labs/eigensdk-go/blob/7c76ea7cdb3bc3c0753de009c01054a8c09446d3/contracts/generate-bindings.sh#L59

shrimalmadhur: c68b91180062786b36dc2566fb7dc393ac673d89

chainio/clients/eigenpod/reader.go
shrimalmadhur: c68b91180062786b36dc2566fb7dc393ac673d89

chainio/clients/eigenpod/generate.sh
shrimalmadhur: c68b91180062786b36dc2566fb7dc393ac673d89

chainio/clients/eigenpod/reader.go
shrimalmadhur: c68b91180062786b36dc2566fb7dc393ac673d89

chainio/clients/eigenpod/reader.go
shrimalmadhur: c68b91180062786b36dc2566fb7dc393ac673d89

PR #318, opened 2024-08-06
feat: add set operator commission bips method

shrimalmadhur:
Fixes # .

### What Changed?
- Add SetOperatorCommissionBips method

### Reviewer Checklist
- [ ] Code is well-documented

### DIFF
--- a/chainio/clients/elcontracts/writer.go
+++ b/chainio/clients/elcontracts/writer.go
@@ -376,3 +376,31 @@ func (w *ChainWriter) ForceDeregisterFromOperatorSets(
 
 	return receipt, nil
 }
+
+func (w *ChainWriter) SetOperatorCommissionBips(
+	ctx context.Context,
+	operatorSet rewardscoordinator.IAVSDirectoryOperatorSet,
+	rewardType uint8,
+	commissionBips uint16,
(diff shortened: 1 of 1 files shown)


chainio/clients/elcontracts/writer.go
```diff
@@ -376,3 +376,31 @@ func (w *ChainWriter) ForceDeregisterFromOperatorSets(
...
+	}
+
+	tx, err := w.rewardsCoordinator.SetOperatorCommissionBips(noSendTxOpts, operatorSet, rewardType, commissionBips)
+	if err != nil {
+		return nil, utils.WrapError("failed to create SetOperatorCommissionBips tx", err)
+	}
+
+	receipt, err := w.txMgr.Send(ctx, tx)
+	if err != nil {
+		return nil, utils.WrapError("failed to send tx", err)
```
samlaf, line 402: We have this pattern in every function. Starting to wonder if we shouldn't refactor it into a function

samlaf (approved):

chainio/clients/elcontracts/writer.go
shrimalmadhur: will update it in consequent PR>

PR #320, opened 2024-08-06
chore: deprecate earning receiver address

shrimalmadhur:
Fixes # .

### What Changed?
- Deprecate Earnings Receiver Address and remove from operator
- a lot of missing fmt and go lines 

### Reviewer Checklist
- [ ] Code is well-documented

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ GO_LINES_IGNORED_DIRS=contracts
 GO_PACKAGES=./chainio/... ./crypto/... ./logging/... \
 	./types/... ./utils/... ./signer/... ./cmd/... \
 	./signerv2/... ./aws/... ./internal/... ./metrics/... \
-	./nodeapi/... ./cmd/... ./services/...
+	./nodeapi/... ./cmd/... ./services/... ./testutils/...
 GO_FOLDERS=$(shell echo ${GO_PACKAGES} | sed -e "s/\.\///g" | sed -e "s/\/\.\.\.//g")
 help:
 	@grep -E '^[a-zA-Z0-9_-]+:.*?## .*$$' $(MAKEFILE_LIST) | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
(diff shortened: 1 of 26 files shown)


Makefile
```diff
@@ -5,7 +5,8 @@
 GO_LINES_IGNORED_DIRS=contracts
 GO_PACKAGES=./chainio/... ./crypto/... ./logging/... \
 	./types/... ./utils/... ./signer/... ./cmd/... \
-	./signerv2/...
+	./signerv2/... ./aws/... ./internal/... ./metrics/... \
+	./nodeapi/... ./cmd/... ./services/... ./testutils/...
```
samlaf, line 9: Is there a way to specify directories to ignore (like .gitignore) instead of ones we want to add? Otherwise we'll keep forgetting to add them here when we add new packages.

chainio/clients/elcontracts/writer.go
```diff
@@ -162,7 +162,10 @@ func (w *ChainWriter) RegisterAsOperator(ctx context.Context, operator types.Ope
 
 	w.logger.Infof("registering operator %s to EigenLayer", operator.Address)
 	opDetails := delegationmanager.IDelegationManagerOperatorDetails{
-		DeprecatedEarningsReceiver: gethcommon.HexToAddress(operator.EarningsReceiverAddress),
+		// Earning receiver has been deprecated but just to make sure we have something in contract
+		// We just force it to be operator address
+		// Any reward related setup is via RewardsCoordinator contract
```
samlaf:
```suggestion
		// Earning receiver has been deprecated so we just use the operator address as a dummy value
		// Any reward related setup is now done via the RewardsCoordinator contract instead
```

samlaf: The comment was a bit confusing to me. I changed it, let me know if it doesn't make sense. If it's clearer, then please also change the other one.

Makefile
shrimalmadhur: yea I checked - this library doesn't do that and requires all pkgs :( . I can write some script to automatically find it I think

chainio/clients/elcontracts/writer.go
shrimalmadhur: 3a5f418fe1bb0ec640e531d17d1d88dbc19aa966

samlaf (approved):

PR #331, opened 2024-08-28
feat: check quorums registered detail

renlulu:
Add a function to query bitmap of quorums and convert it into human-readable format which is boolean array.
### Reviewer Checklist
- [ ] Code is well-documented

### DIFF
--- a/chainio/clients/avsregistry/reader.go
+++ b/chainio/clients/avsregistry/reader.go
@@ -358,18 +358,24 @@ func (r *ChainReader) GetOperatorFromId(
 func (r *ChainReader) QueryRegistrationDetail(
 	opts *bind.CallOpts,
 	operatorAddress common.Address,
-) (map[int]bool, error) {
+) ([]bool, error) {
 	operatorId, err := r.GetOperatorId(opts, operatorAddress)
 	if err != nil {
 		return nil, err
 	}
-	bitmap, err := r.registryCoordinator.GetCurrentQuorumBitmap(opts, operatorId)
+	value, err := r.registryCoordinator.GetCurrentQuorumBitmap(opts, operatorId)
 	if err != nil {
 		return nil, err
 	}
-	quorums := make(map[int]bool)
-	for i, b := range bitmap.Bits() {
-		quorums[i] = b == 1
+	numBits := value.BitLen()
+	quorums := make([]bool, numBits)
+	for i := 0; i < numBits; i++ {
+		// Check if the ith bit is set
+		if value.Int64()&(1<<i) != 0 {
+			quorums[i] = true
+		} else {
+			quorums[i] = false
+		}
 	}
 	return quorums, nil
 }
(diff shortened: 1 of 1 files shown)


chainio/clients/avsregistry/reader.go
```diff
@@ -355,6 +355,31 @@ func (r *ChainReader) GetOperatorFromId(
...
+	operatorId, err := r.GetOperatorId(opts, operatorAddress)
+	if err != nil {
+		return nil, err
+	}
+	value, err := r.registryCoordinator.GetCurrentQuorumBitmap(opts, operatorId)
+	if err != nil {
+		return nil, err
+	}
+	numBits := value.BitLen()
+	quorums := make([]bool, numBits)
```
samlaf: How are you using this quorums array? Wouldn't you rather have a map instead?

renlulu: Hi, initially i thought we could use index of the array as the quorum index, but yes i have changed it to use a map for  more easier to understand

chainio/clients/avsregistry/reader.go
```diff
@@ -355,6 +355,26 @@ func (r *ChainReader) GetOperatorFromId(
...
+		return nil, err
+	}
+	value, err := r.registryCoordinator.GetCurrentQuorumBitmap(opts, operatorId)
+	if err != nil {
+		return nil, err
+	}
+	numBits := value.BitLen()
+	quorums := make(map[int]bool, numBits)
+	for i := 0; i < numBits; i++ {
+		quorums[i] = value.Int64()&(1<<i) != 0
```
samlaf: Haven’t used quorum bitmaps in a long while. Are you sure this is how they work? Aka the quorums are laid out from right to left and not the other way around?

chainio/clients/avsregistry/reader.go
samlaf: Think you should only be inputting the non zero quorums into the map? Aka treating it as a set instead of a map. That way you would iterate only on quorums that are nonzero.

chainio/clients/avsregistry/reader.go
renlulu: Yep, i have tested 2 cases:
1. opt-in into 3 quorums: 0,1,2 then eject operator from 1, the contract returns 5 whose bit format is: 1,0,1
2. opt-in into 3 quorums: 0,1,2 them eject operator from 0, the contract returns 6 whose bit format is: 1,1,0
image
image

chainio/clients/avsregistry/reader.go
renlulu: I think it is better to show zero quorums as well? so one use this function can easily get:
1. which quorum(s) the operator is in already
2. which quorum(s) the operator is not in

so he can adjust the strategy accordingly, i.e. re-join the missing quorum again

chainio/clients/avsregistry/reader.go
samlaf: Awesome will trust you on this one then

chainio/clients/avsregistry/reader.go
samlaf: I see. In that case don’t think map is needed and we can revert to array then haha?

renlulu: ok

renlulu: reverted to use array

samlaf (approved):

PR #332, opened 2024-08-29
refactor: handle zero bitmap

renlulu:
A followup PR of https://github.com/Layr-Labs/eigensdk-go/pull/331. The functions works well with our system internally, but i just realized that if the bitmap is 0, the result array would be empty too. So is it better that we populate all existing quorums with default value (which is false in this case) to the result to be more consistent. Also wrap some errors this time.

### Reviewer Checklist
- [ ] Code is well-documented

### DIFF
--- a/chainio/clients/avsregistry/reader.go
+++ b/chainio/clients/avsregistry/reader.go
@@ -361,16 +361,25 @@ func (r *ChainReader) QueryRegistrationDetail(
 ) ([]bool, error) {
 	operatorId, err := r.GetOperatorId(opts, operatorAddress)
 	if err != nil {
-		return nil, err
+		return nil, utils.WrapError("Failed to get operator id", err)
 	}
 	value, err := r.registryCoordinator.GetCurrentQuorumBitmap(opts, operatorId)
 	if err != nil {
-		return nil, err
+		return nil, utils.WrapError("Failed to get operator quorums", err)
 	}
 	numBits := value.BitLen()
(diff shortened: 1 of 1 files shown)


chainio/clients/avsregistry/reader.go
```diff
@@ -361,17 +361,26 @@ func (r *ChainReader) QueryRegistrationDetail(
...
 	if err != nil {
-		return nil, err
+		return nil, utils.WrapError("Failed to get operator quorums", err)
 	}
 	numBits := value.BitLen()
 	var quorums []bool
 	for i := 0; i < numBits; i++ {
 		quorums = append(quorums, value.Int64()&(1<<i) != 0)
 	}
+	if len(quorums) == 0 {
```
samlaf, line 375: can you add a comment explaining when/why this case is possible? When does GetCurrentQuorumBitmap return a value of 0 bitlen?

renlulu: I think if operator has opt out from all quorums (or been ejected), the contract could return 0, you can see here:

renlulu: I think if operator has opt out from all quorums (or been ejected), the contract could return 0, you can see here:
image

samlaf: meant in the code, because it's surprising when reading the code why it could ever be 0.

samlaf (approved):

PR #334, opened 2024-09-04
fest: support start/stop blocks for querying operator info(pubkey and socket)

### LINKED ISSUE #259: Add an option for ```NewOperatorsInfoServiceInMemory``` to start scanning from a given block
```NewOperatorsInfoServiceInMemory``` takes too much time to initialize.

It could start scanning the blockchain from some arbitrary block. ```QueryExistingRegisteredOperatorPubKeys``` already provides that option.

### PR DESCRIPTION
**Pull request by renlulu**
Fixes #259 and #312.

### Reviewer Checklist
- [ ] Code is well-documented

### DIFF
--- a/services/operatorsinfo/operatorsinfo_inmemory.go
+++ b/services/operatorsinfo/operatorsinfo_inmemory.go
@@ -78,6 +78,11 @@ type resp struct {
 	operatorExists bool
 }
 
+type Opts struct {
+	StartBlock *big.Int
+	StopBlock  *big.Int
+}
+
 var _ OperatorsInfoService = (*OperatorsInfoServiceInMemory)(nil)
 
 // NewOperatorsInfoServiceInMemory constructs a OperatorsInfoServiceInMemory and starts it in a goroutine.
@@ -92,8 +97,7 @@ func NewOperatorsInfoServiceInMemory(
 	avsRegistrySubscriber avsRegistrySubscriber,
 	avsRegistryReader avsRegistryReader,
 	logFilterQueryBlockRange *big.Int,
-	startBlock *big.Int,
-	stopBlock *big.Int,
+	opts *Opts,
 	logger logging.Logger,
 ) *OperatorsInfoServiceInMemory {
 	queryC := make(chan query)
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by samlaf — COMMENTED** [review]
This looks fine to me. @shrimalmadhur do you want to start enforcing opts arguments so that we don't have breaking changes like this all the time?

**Comment by shrimalmadhur** [comment]
> This looks fine to me. @shrimalmadhur do you want to start enforcing opts arguments so that we don't have breaking changes like this all the time?

yea this is getting very ugly. these seems like optional params to me - if we don't provide them, it takes default values. We should add extensible options without breaking the method arguments.

**Comment by samlaf** [comment]
Ok @renlulu synced with @shrimalmadhur , we're trying to enforce opts params going forward, such that we don't break client code everytime we add new params to a function.

Would you mind refactoring the function signatures everywhere to use a params struct instead that would for now contain the startBlock and endBlock?
```
type Opts struct {
   StartBlock *big.Int
   StopBlock *big.Int
}
```

Not sure about the exact name, `Opts` or `Params` would work, prob with a prefix related to name of function in case there's a few of them in a given module.

**Comment by renlulu** [comment]
> Ok @renlulu synced with @shrimalmadhur , we're trying to enforce opts params going forward, such that we don't break client code everytime we add new params to a function.
> 
> Would you mind refactoring the function signatures everywhere to use a params struct instead that would for now contain the startBlock and endBlock?
> 
> ```
> type Opts struct {
>    StartBlock *big.Int
>    StopBlock *big.Int
> }
> ```
> 
> Not sure about the exact name, `Opts` or `Params` would work, prob with a prefix related to name of function in case there's a few of them in a given module.

Ok, make sense, let me refactor it.

File: services/operatorsinfo/operatorsinfo_inmemory.go [hunk 1]
```diff
@@ -298,8 +300,8 @@ func (ops *OperatorsInfoServiceInMemory) queryPastRegisteredOperatorEventsAndFil
 	go func() {
 		alreadyRegisteredOperatorAddrs, alreadyRegisteredOperatorPubkeys, pubkeysErr = ops.avsRegistryReader.QueryExistingRegisteredOperatorPubKeys(
 			ctx,
-			startBlock,
-			stopBlock,
+			opts.StartBlock,
+			opts.StopBlock,
```
samlaf, line 304: this will panic if `nil` is passed as opts. Can we change opts to be `Opts` instead of `*Opts`?

samlaf: in every function

samlaf (approved):
Thank you!

PR #336, opened 2024-09-09
chore: test util funcs

estensen:
Test coverage is pretty low. First stab at improving it.

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -40,6 +40,21 @@ var (
 	// We do NOT want to support formats like SVG since they can be used for javascript injection
 	// If we get pushback on only supporting png, we can support jpg, jpeg, gif, etc. later
 	ImageExtensions = []string{".png"}
+
+	// Regular expression to ethereum address
+	ethAddrPattern = regexp.MustCompile("^0x[0-9a-fA-F]{40}$")
+
+	// Regular expression to validate text
+	textPattern = regexp.MustCompile(TextRegex)
+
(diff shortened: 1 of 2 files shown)


samlaf (commented):
LGTM. @estensen is this ready for review or were you planning on doing more work here?

@shrimalmadhur wdyt about breaking up this utils dir into more useful named modules (like `utils/url_validation` or `utils/validation`)? Also potentially moving this into `internal/` to make sure other people don't rely on these functions, such that we can keep refactoring them and moving them around. Unless the CLI is using them?

shrimalmadhur:
> LGTM. @estensen is this ready for review or were you planning on doing more work here?
> 
> @shrimalmadhur wdyt about breaking up this utils dir into more useful named modules (like `utils/url_validation` or `utils/validation`)? Also potentially moving this into `internal/` to make sure other people don't rely on these functions, such that we can keep refactoring them and moving them around. Unless the CLI is using them?

Yes. I think that makes sense. I think CLI should make it's own utils and not rely on sdk for internal utilities. I have changed some already but if I finding anything breaking I will change it in CLI. so ya let's move internal utils to internal and break them down.

estensen:
> LGTM. @estensen is this ready for review or were you planning on doing more work here?

Trying to keep diffs small, so this is it for now. Was planning on continuing to have at least some tests for all functions here to get better test coverage.

samlaf (approved):

PR #338, opened 2024-09-13
fix: rpc request duration in seconds

### LINKED ISSUE #333: The `eigen_rpc_request_duration_seconds` need change `_seconds` suffix of the metric name
**Describe the bug**

Noticed that the `eigen_rpc_request_duration_seconds` histogram metric seems to record abnormal values - it looks like they are expressed in nanoseconds which contradicts the `_second` suffix of the metric name

**Expected behavior**

For this request duration, we need change it 's name.

**Screenshots**

**OS details**

**Additional context**

renlulu:
Fixes #333  .

### Reviewer Checklist
- [ ] Code is well-documented


samlaf (approved):
Thanks for this!

PR #340, opened 2024-09-18
chore: test more utils funcs

estensen:
Continuation of https://github.com/Layr-Labs/eigensdk-go/pull/336


utils/utils.go
```diff
@@ -105,6 +105,7 @@ func EcdsaPrivateKeyToAddress(privateKey *ecdsa.PrivateKey) (gethcommon.Address,
 	return crypto.PubkeyToAddress(*publicKeyECDSA), nil
 }
 
+// RoundUpDivideBig divides two positive big.Int numbers and rounds up the result.
```
estensen, line 65: Not sure what this is used for, but it currently doesn't handle negative numbers

utils/utils.go
```diff
@@ -118,14 +119,13 @@ func IsValidEthereumAddress(address string) bool {
...
 }
 
-func ReadPublicURL(url string) ([]byte, error) {
-	// allow no redirects
-	httpClient := http.Client{
-		CheckRedirect: func(req *http.Request, via []*http.Request) error {
-			return http.ErrUseLastResponse
-		},
-		Timeout: 3 * time.Second,
+func ReadPublicURL(url string, httpClient *http.Client) ([]byte, error) {
```
estensen: Had to change the signature to make it possible to mock. IMO we shouldn't do real HTTP reqs from tests.

utils/utils_test.go
```diff
@@ -1,51 +1,470 @@
...
+			yamlContent: `
+name: Alice
+`,
+			expected: Config{Name: "Alice", Age: 0},
+		},
+		{
+			name:        "Empty YAML",
+			yamlContent: ``,
+			expected:    Config{},
+		},
```
estensen: Not able to test with invalid values here or for `TestReadJsonConfig` since it panics. Can rewrite to return an error instead of panic so this can be tested. Consider using recover, but `log.Fatal` creates an unrecoverable panic.

utils/utils.go
MegaRedHand: This function is unused. I'm guessing it was not supposed to be public, so I marked it as deprecated.

utils/utils.go
MegaRedHand: I agree that doing HTTP requests to a remote server on tests is not a good idea. In this case the server is started by the same test, so it shouldn't give any problems. Also, adding the client parameter breaks the interface, which we can't do.

MegaRedHand:
Thanks for your contribution @estensen! Since the PR is almost a year old I took the liberty of pushing some edits to it:

- Reverted the changes to the `utils` module to avoid breaking the public API
- Removed the tests for `ReadFile`, `ReadYamlConfig`, and `ReadYamlConfig`, since those are no longer in the SDK
- Marked `RoundUpDivideBig` as deprecated, and added a comment about negative numbers not working.

MegaRedHand (approved):

PR #342, opened 2024-09-18
feat: slashing

shrimalmadhur:
[DO NOT MERGE]

Fixes # .

### What Changed?
- Slashing and Operator Set release

### Reviewer Checklist
- [ ] Code is well-documented


chainio/clients/elcontracts/writer.go
```diff
@@ -3,7 +3,7 @@ package elcontracts
 import (
 	"context"
 	"errors"
-
+	permissioncontroller "github.com/Layr-Labs/eigensdk-go/contracts/bindings/PermissionController"
```
shrimalmadhur: move this down to where other contracts are defined

chainio/clients/elcontracts/writer.go
```diff
@@ -545,3 +550,49 @@ func (w *ChainWriter) RegisterForOperatorSets(
 
 	return receipt, nil
 }
+
+func (w *ChainWriter) RemovePermission(ctx context.Context, request RemovePermissionRequest) error {
```
shrimalmadhur: you need to send receipt back - so that you can populate transaction details on CLI. check the other write methods.

chainio/clients/elcontracts/writer.go
```diff
@@ -545,3 +550,49 @@ func (w *ChainWriter) RegisterForOperatorSets(
...
+		request.Selector,
+	)
+	if err != nil {
+		return errors.New("call to permission controller failed: " + err.Error())
+	}
+	_, err = w.txMgr.Send(ctx, tx, request.WaitForReceipt)
+	return err
+}
+
+func (w *ChainWriter) SetPermission(ctx context.Context, request SetPermissionRequest) error {
```
shrimalmadhur: you need to send receipt back - so that you can populate transaction details on CLI. check the other write methods.

bdchatham (dismissed):

chainio/clients/elcontracts/bindings.go
```diff
@@ -143,17 +146,15 @@ func NewEigenlayerContractBindings(
 		return nil, utils.WrapError("Failed to fetch StrategyManager contract", err)
 	}
 
-	avsDirectory, err := avsdirectory.NewContractIAVSDirectory(avsDirectoryAddr, ethclient)
+	avsDirectory, err := avsdirectory.NewContractAVSDirectory(avsDirectoryAddr, ethclient)
 	if err != nil {
 		return nil, utils.WrapError("Failed to fetch AVSDirectory contract", err)
 	}
 
 	return &ContractBindings{
```
bdchatham, line 158: Is there a reason on why we're not including the new bindings in this? I had omitted permission controller because I saw you hadn't added allocation manager. I know it's deprecated, but it's not clear why we shouldn't make it work, especially when it's this easy.

cmd/egnaddrs/main.go
```diff
@@ -210,17 +210,17 @@ func getEigenlayerContractAddrs(
...
 	}
-	strategyManagerAddr, err := delegationManagerC.StrategyManager(&bind.CallOpts{})
+	allocationManagerAddr, err := delegationManagerC.AllocationManager(&bind.CallOpts{})
 	if err != nil {
 		return nil, err
 	}
 
 	addrsDict := map[string]string{
-		"slasher":            slasherAddr.Hex(),
+		"allocation-manager": allocationManagerAddr.Hex(),
```
bdchatham, line 223: Need to add permission-controller.

contracts/script/parsers/EigenlayerContractsParser.sol
```diff
@@ -1,28 +1,29 @@
...
 import {StrategyBaseTVLLimits} from "eigenlayer-contracts/src/contracts/strategies/StrategyBaseTVLLimits.sol";
 import {IRewardsCoordinator} from "eigenlayer-contracts/src/contracts/interfaces/IRewardsCoordinator.sol";
+import {IAllocationManager} from "eigenlayer-contracts/src/contracts/interfaces/IAllocationManager.sol";
 
 import {ConfigsReadWriter} from "./ConfigsReadWriter.sol";
 import "forge-std/StdJson.sol";
+import {IAllocationManager} from "../../lib/eigenlayer-middleware/lib/eigenlayer-contracts/src/contracts/interfaces/IAllocationManager.sol";
 
-struct EigenlayerContracts {
+    struct EigenlayerContracts {
```
bdchatham, line 18: What is the used for? I expect PermissionController is needed?

chainio/clients/elcontracts/bindings.go
shrimalmadhur: yea because we are gonna remove it soon. I think there's already a PR which removes it

contracts/script/parsers/EigenlayerContractsParser.sol
shrimalmadhur: this is needed to parse the contracts for deploying AVS to the anvil deployment. deploying an AVS won't need PermissionController so we might not need it.

cmd/egnaddrs/main.go
shrimalmadhur: can we fetch permission controller address from any other contract? I am assuming no. then we don't need to add that as you explicitly need to know that. over time with zeus this egnaddrs will be deprecated

bdchatham (dismissed):

chainio/clients/eigenpod/generate.sh
```diff
@@ -16,7 +16,7 @@ cleanup_bindings_dir() {
 
 clone() {
   echo "Cloning the EigenLayer contracts repository"
-  git clone -b feat/partial-withdrawal-batching --depth=1 git@github.com:Layr-Labs/eigenlayer-contracts.git "${TMP_DIR}"
+  git clone -b slashing-magnitudes --depth=1 git@github.com:Layr-Labs/eigenlayer-contracts.git "${TMP_DIR}"
```
afkbyte, line 19: the eigenpod specific bindings generation is probably not necessary anymore

chainio/clients/elcontracts/bindings.go
```diff
@@ -58,29 +61,37 @@ func NewBindingsFromConfig(
...
+			return nil, utils.WrapError("Failed to fetch AllocationManager contract", err)
 		}
 	}
 
+	permissionController, err = permissioncontroller.NewContractPermissionController(
+		cfg.PermissionsControllerAddress,
+		client,
+	)
+	if err != nil {
+		return nil, utils.WrapError("Failed to fetch RewardsCoordinator contract", err)
```
afkbyte:
```suggestion
		return nil, utils.WrapError("Failed to fetch PermissionController contract", err)
```

chainio/clients/elcontracts/writer.go
```diff
@@ -442,3 +451,325 @@ func (w *ChainWriter) ProcessClaims(
...
 }
+
+func (w *ChainWriter) ForceDeregisterFromOperatorSets(
+	ctx context.Context,
+	operator gethcommon.Address,
+	avs gethcommon.Address,
+	operatorSetIds []uint32,
+	waitForReceipt bool,
+) (*gethtypes.Receipt, error) {
+	if w.avsDirectory == nil {
```
afkbyte: I think we should check for the `allocationManager` here rather than the `avsDirectory`

chainio/clients/elcontracts/writer.go
```diff
@@ -442,3 +451,325 @@ func (w *ChainWriter) ProcessClaims(
...
+	}
+	receipt, err := w.txMgr.Send(ctx, tx, waitForReceipt)
+	if err != nil {
+		return nil, utils.WrapError("failed to send tx", err)
+	}
+
+	return receipt, nil
+}
+
+func (w *ChainWriter) DeregisterFromOperatorSets(
```
afkbyte, line 547: I think we still need to support the legacy registration and deregistration flows in the sdk see: https://github.com/Layr-Labs/eigenlayer-contracts/blob/813f52cd125b9fb46f69fb5679a930b030187a2b/src/contracts/core/AVSDirectory.sol#L71 as they'll be used to migrate iirc

contracts/anvil/utils.sh
```diff
@@ -27,13 +27,15 @@ start_anvil_docker() {
     DUMP_STATE_VOLUME_DOCKER_ARG=$([[ -z $DUMP_STATE_FILE ]] && echo "" || echo "-v $DUMP_STATE_FILE:/dump-state.json")
     LOAD_STATE_ANVIL_ARG=$([[ -z $LOAD_STATE_FILE ]] && echo "" || echo "--load-state /load-state.json")
     DUMP_STATE_ANVIL_ARG=$([[ -z $DUMP_STATE_FILE ]] && echo "" || echo "--dump-state /dump-state.json")
+    # temp fix since one of the middleware contract's is greater than default code size limit
+    CODE_SIZE_ARG="--code-size-limit 30000"
```
afkbyte: iirc the issue is with the new `DeployMockAvsRegistries.s.sol`, I refactored it [here](https://github.com/Layr-Labs/eigensdk-go/blob/2e5ef1f871760bbe2b8019d39edc75f23d406e53/contracts/script/DeployMockAvsRegistries.s.sol) if you want to bring it in

afkbyte:
want to also note that the `nightly-3abac322efdb69e27b6fe8748b72754ae878f64d` image we're pulling of anvil no longer exists - we should use the latest stable https://github.com/foundry-rs/foundry/pkgs/container/foundry/326174124?tag=stable (image should be "ghcr.io/foundry-rs/foundry:stable@sha256:daeeaaf4383ee0cbfc9f31f079a04ffb0123e49e5f67f2a20b5ce1ac1959a4d6")

chainio/clients/elcontracts/writer.go
MegaRedHand: Can't we keep the check for any of them? Like

```go
if w.avsDirectory == nil || w.allocationManager == nil {
```

That way users have the option to keep using the old workflow with the new version of the SDK

chainio/clients/eigenpod/generate.sh
shrimalmadhur: Yea I will clean that up later. we did it because of incompatibility.

contracts/anvil/utils.sh
shrimalmadhur: hmm interesting - from what I remember it was one of the middleware contracts but will remove it now and bring in the refactor

chainio/clients/elcontracts/bindings.go
shrimalmadhur: good eyes!

chainio/clients/elcontracts/writer.go
shrimalmadhur: that method in is called by AVS service manager iirc right? I think we never had those in core bindings as they would never be called directly.

contracts/anvil/utils.sh
shrimalmadhur: done! thanks for the refactor!

chainio/clients/elcontracts/writer.go
afkbyte: I think we maintained it in the core bindings (at least we did with register as operator) bc not all avs were using our service manager

https://github.com/Layr-Labs/eigensdk-go/blob/2c554f1a02be8e59be2e712a35551e297700d89a/chainio/clients/elcontracts/writer.go#L157

shrimalmadhur: we still have that - https://github.com/Layr-Labs/eigensdk-go/blob/859288176f423c69637a21a6ef50fabf73f7475a/chainio/clients/elcontracts/writer.go#L167 
but this is for register to eigenlayer core, it has no relation with AVS registration per se.

afkbyte: ahh ok my bad!

afkbyte (approved):
Lgtm!

PR #343, opened 2024-09-18
fix: race condition when querying registered operators data

ricomateo:
### What Changed?
This PR adds a small fix for a race condition found when querying registered operators data.

The functions `QueryExistingRegisteredOperatorPubKeys` and `QueryExistingRegisteredOperatorSockets` both run in parallel and they read and mutate the same variable `startBlock`. This causes a race condition because `startBlock` is a `*big.Int`.

The solution is to clone the `startBlock` into a new variable for each function.

A better solution would be to change `QueryExistingRegisteredOperatorPubKeys` and `QueryExistingRegisteredOperatorSockets` function signatures to receive `startBlock` as primitive `int64` variable instead of `*big.Int` and do the conversion to `*big.Int` inside of the functions.

### Reviewer Checklist
- [ ] Code is well-documented

### DIFF
--- a/chainio/clients/avsregistry/reader.go
+++ b/chainio/clients/avsregistry/reader.go
@@ -426,11 +426,10 @@ func (r *ChainReader) QueryExistingRegisteredOperatorPubKeys(
 	if blockRange == nil {
 		blockRange = DefaultQueryBlockRange
 	}
-	startBlock = new(big.Int).Set(startBlock)
 
 	operatorAddresses := make([]types.OperatorAddr, 0)
 	operatorPubkeys := make([]types.OperatorPubkeys, 0)
-	for i := startBlock; i.Cmp(stopBlock) <= 0; i.Add(i, blockRange) {
+	for i := new(big.Int).Set(startBlock); i.Cmp(stopBlock) <= 0; i.Add(i, blockRange) {
 		// Subtract 1 since FilterQuery is inclusive
 		toBlock := big.NewInt(0).Add(i, big.NewInt(0).Sub(blockRange, big.NewInt(1)))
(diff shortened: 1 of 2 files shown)


chainio/clients/avsregistry/reader.go
```diff
@@ -429,7 +429,7 @@ func (r *ChainReader) QueryExistingRegisteredOperatorPubKeys(
 
 	operatorAddresses := make([]types.OperatorAddr, 0)
 	operatorPubkeys := make([]types.OperatorPubkeys, 0)
-	for i := startBlock; i.Cmp(stopBlock) <= 0; i.Add(i, blockRange) {
+	for i := new(big.Int).Set(startBlock); i.Cmp(stopBlock) <= 0; i.Add(i, blockRange) {
```
samlaf, line 437:
```suggestion
        // QueryExistingRegisteredOperatorPubKeys and QueryExistingRegisteredOperatorSockets both run in parallel and they read and mutate the same variable startBlock, so we clone it to prevent the race condition
        // TODO: we might want to eventually change the function signature to pass a uint, but that would be a breaking change
	for i := new(big.Int).Set(startBlock); i.Cmp(stopBlock) <= 0; i.Add(i, blockRange) {
```

samlaf (commented):
Thanks for this fix :) Let's just add a comment to explain why we are doing this.

chainio/clients/avsregistry/reader.go
ricomateo: Done!

samlaf (approved):
LGTM thanks! operator unit test is failing, we know why. Will push a fix soon and then you'll just have to rebase on top before merging this PR.

PR #344, opened 2024-09-23
Avoid panic in ReadStringFromSecretManager in secretmanager.go

pablodeymo:
Fixes # .

### What Changed?
Avoid panicking in secretmanager.go, returning the error instead of doing a `Fatal`.

### Reviewer Checklist
- [ ] Code is well-documented

### DIFF
--- a/aws/secretmanager/secretmanager.go
+++ b/aws/secretmanager/secretmanager.go
@@ -3,6 +3,7 @@ package secretmanager
 import (
 	"context"
 
+	"github.com/Layr-Labs/eigensdk-go/utils"
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/config"
 	"github.com/aws/aws-sdk-go-v2/service/secretsmanager"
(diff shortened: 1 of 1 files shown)


aws/secretmanager/secretmanager.go
```diff
@@ -12,7 +11,7 @@ import (
 func ReadStringFromSecretManager(ctx context.Context, secretName, region string) (string, error) {
 	config, err := config.LoadDefaultConfig(ctx, config.WithRegion(region))
 	if err != nil {
-		log.Fatal(err)
+		return "", err
```
samlaf:
```suggestion
		return "", errors.Wrap("error while loading default config", err)
```

samlaf: nit: we have a wrap errors function somewhere in utils. Would you mind wrapping with that to give error context?

samlaf (commented):
Agree with this, would just prefer errors to be wrapped everywhere.

aws/secretmanager/secretmanager.go
pablodeymo: done! thanks

shrimalmadhur (approved):

PR #345, opened 2024-09-23
Prevent Duplicate Signature Processing in BLS Aggregator

afkbyte:
Fixes [AVS-1053]( https://linear.app/eigenlabs/issue/AVS-1053/avs-devex-prevent-duplicate-signature-processing-in-bls-aggregator)

### What Changed?
Modified `singleTaskAggregatorGoroutineFunc` in blsagg/bls_aggregator_service.go
- Added explicit duplicate signature check before signature verification
- Moved task response hashing earlier in the processing flow
- Added unit test 

### Reviewer Checklist
- [ ] Code is well-documented

### DIFF
--- a/services/bls_aggregation/blsagg.go
+++ b/services/bls_aggregation/blsagg.go
@@ -336,12 +336,6 @@ func (a *BlsAggregatorService) singleTaskAggregatorGoroutineFunc(
 				signedTaskResponseDigest,
 			)
 
-			err := a.verifySignature(taskIndex, signedTaskResponseDigest, operatorsAvsStateDict)
-			signedTaskResponseDigest.SignatureVerificationErrorC <- err
-			if err != nil {
-				continue
-			}
-
 			// compute the taskResponseDigest using the hash function
 			taskResponseDigest, err := a.hashFunction(signedTaskResponseDigest.TaskResponse)
 			if err != nil {
@@ -350,8 +344,23 @@ func (a *BlsAggregatorService) singleTaskAggregatorGoroutineFunc(
 				// happens..
 				continue
 			}
-			// after verifying signature we aggregate its sig and pubkey, and update the signed stake amount
+
+			// check if the operator has already signed for this digest
 			digestAggregatedOperators, ok := aggregatedOperatorsDict[taskResponseDigest]
+			if ok {
+				if digestAggregatedOperators.signersOperatorIdsSet[signedTaskResponseDigest.OperatorId] {
+					signedTaskResponseDigest.SignatureVerificationErrorC <- fmt.Errorf("duplicate signature from operator %x for task %d", signedTaskResponseDigest.OperatorId, taskIndex)
+					continue
+				}
+			}
+
+			err = a.verifySignature(taskIndex, signedTaskResponseDigest, operatorsAvsStateDict)
+			signedTaskResponseDigest.SignatureVerificationErrorC <- err
+			if err != nil {
+				continue
+			}
+
+			// after verifying signature we aggregate its sig and pubkey, and update the signed stake amount
 			if !ok {
 				// first operator to sign on this digest
 				digestAggregatedOperators = aggregatedOperators{
--- a/services/bls_aggregation/blsagg.go
+++ b/services/bls_aggregation/blsagg.go
@@ -336,29 +336,12 @@ func (a *BlsAggregatorService) singleTaskAggregatorGoroutineFunc(
 				signedTaskResponseDigest,
 			)
 
-			err := a.verifySignature(taskIndex, signedTaskResponseDigest, operatorsAvsStateDict)
-			if err != nil {
-				a.logger.Info(
-					"Signature verification failed",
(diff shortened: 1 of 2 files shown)


services/bls_aggregation/blsagg.go
```diff
@@ -350,8 +344,28 @@ func (a *BlsAggregatorService) singleTaskAggregatorGoroutineFunc(
 				// happens..
 				continue
 			}
-			// after verifying signature we aggregate its sig and pubkey, and update the signed stake amount
+
+			// check if the operator has already signed for this digest
 			digestAggregatedOperators, ok := aggregatedOperatorsDict[taskResponseDigest]
+			if ok {
+				if digestAggregatedOperators.signersOperatorIdsSet[signedTaskResponseDigest.OperatorId] {
+					a.logger.Warn(
```
samlaf:
```suggestion
					a.logger.Info(
```

samlaf: Let's keep warning for things that shouldn't happen but that we should pay attention to, see https://stackoverflow.com/a/2031209/4971151

duplicate signature isn't interesting, unless we receive 10k of them say

services/bls_aggregation/blsagg.go
```diff
@@ -350,8 +344,28 @@ func (a *BlsAggregatorService) singleTaskAggregatorGoroutineFunc(
 				// happens..
 				continue
 			}
-			// after verifying signature we aggregate its sig and pubkey, and update the signed stake amount
+
+			// check if the operator has already signed for this digest
 			digestAggregatedOperators, ok := aggregatedOperatorsDict[taskResponseDigest]
+			if ok {
```
samlaf, line 350: there's already an `if ok {} else {}` conditional below. Probably just want to put everything in here in the else clause below?

services/bls_aggregation/blsagg.go
```diff
@@ -350,8 +344,28 @@ func (a *BlsAggregatorService) singleTaskAggregatorGoroutineFunc(
...
+						"operatorId", fmt.Sprintf("%x", signedTaskResponseDigest.OperatorId),
+						"taskIndex", taskIndex,
+					)
+					signedTaskResponseDigest.SignatureVerificationErrorC <- fmt.Errorf("duplicate signature from operator %x for task %d", signedTaskResponseDigest.OperatorId, taskIndex)
+					continue
+				}
+			}
+
+			err = a.verifySignature(taskIndex, signedTaskResponseDigest, operatorsAvsStateDict)
+			signedTaskResponseDigest.SignatureVerificationErrorC <- err
```
samlaf, line 365:
```suggestion
                         // return the err (or nil) to the operator, and then proceed to do aggregation logic asynchronously (when no error)
			signedTaskResponseDigest.SignatureVerificationErrorC <- err
```

samlaf: Can we add a comment like this? Even I was confused reading my old code as to why I was returning a potentially nil error to the sender here on first reading haha

samlaf (commented):
Nice fix! Few nits to make the code more easily readable

services/bls_aggregation/blsagg.go
MegaRedHand: Wouldn't this mess with the error handling? Between this `if ok` and the other `if !ok {} else {}` we unconditionally send the result to the error channel. We could move the `if/else` here, but that would increase the amount of computation before sending a result.

services/bls_aggregation/blsagg.go
afkbyte: https://github.com/Layr-Labs/eigensdk-go/pull/345/commits/1fa0bb48927c69a7557df88627248062d6cfaa4f maybe adaptive logging is something we should consider as an enhancement down the line, there's probably a few places it would be useful

services/bls_aggregation/blsagg.go
afkbyte: https://github.com/Layr-Labs/eigensdk-go/pull/345/commits/fb2628d6616b8d40ea1954f9936003578788badc

services/bls_aggregation/blsagg.go
afkbyte: I guess the performance impact would be minimal - I'll switch to the more readable approach with the signature verification before

afkbyte: 7627a5eec95d20ce0c9538fedfde374ab990696d

afkbyte: scratch that - there's not really a way to move it to the  if/else without restructuring the code a decent amount because of how the error channels are consumed

samlaf: Ah my bad... ugly formatting on github made it look like all the logic was part of the same if ok {} conditional. Sorry for making you waste time here @afkbyte :\

samlaf (approved):
LGTM thanks! operator unit test is failing, we know why. Will push a fix soon and then you'll just have to rebase on top before merging this PR.

PR #350, opened 2024-09-24
test: cleanup operator_test.go to only make local calls (no internet dependencies)

samlaf:
The tests were failing this morning when `http://example.com` was returning 500 errors temporarily. We shouldn't depend on any external servers for tests, so this PR is a wip DRAFT for this goal.

Super ugly right now, but at least it's working. Just need to clean this up.


types/operator_test.go
```diff
@@ -1,15 +1,96 @@
...
+			w.WriteHeader(http.StatusNotFound)
+			return
+		}
+		if r.URL.Path == "/operator.png" {
+			w.Header().Set("Content-Type", "image/png")
+			http.ServeContent(w, r, "operator.png", time.Now(), bytes.NewReader(operatorPngBuffer.Bytes()))
+			return
+		}
+		marshalledOperatorMetadata, err := json.Marshal(operatorMetadata)
+		require.NoError(t, err)
```
shrimalmadhur, line 46: I forgot what was diff between `require` and `assert`?

shrimalmadhur (approved):

PR #359, opened 2024-09-27
ChainID method in HttpBackend and WsBackend interfaces

pablodeymo:
For the point of view of the user of the SDK, it's better to have the ChainID method in HttpBackend and WsBackend interfaces.

The idea behind instrumentedClient is to wrap a Eth Client, so the user can use both clients the same way.
However, there are some methods missing from the common interfaces.

It would be good to review if there are other methods missing on that interfaces, for future PRs.

### Reviewer Checklist
- [ ] Code is well-documented


samlaf:
Hey @pablodeymo , if chainid is not used anywhere, why add it? I prefer keeping interfaces to only what is needed, not bloating them just in case.

pablodeymo:
> Hey @pablodeymo , if chainid is not used anywhere, why add it? I prefer keeping interfaces to only what is needed, not bloating them just in case.

This method is used in the Incredible Squaring example ([here](https://github.com/Layr-Labs/incredible-squaring-avs/blob/20f4792c3521399fcfb5365386c2ea7b2e44cd9d/operator/operator.go#L129)). Since its removal, it's difficult to have optionally-enabled metrics since restricting the type to the `xxBackend` interface would also restrict the functionality unnecessarily.

samlaf:
@pablodeymo I see. I guess it’s fine to add, but this makes me think we haven’t found the right abstractions yes :/
Perhaps inc-sq should define the interface it needs in this case. Since we return structs in the constructors everything should work.

PR #386, opened 2024-10-28
Add BLS Aggregation window waiting after quorum

### LINKED ISSUE #300: Feat(blsagg): use block-based expiry instead of time-based

bls signature verification gas costs are linear in the # of non signers, so the fewer nonsigners, the cheaper the verification gas costs. Right now the API for the blsagg service to return an aggregate signature has 2 dofs:
- minimum signed stake threshold (60%, 80%, etc.)
- time to expiry  (see this [TODO](https://github.com/Layr-Labs/eigensdk-go/blob/4b1bb2a86e2acc0de02952b5a20de1aa5eaaf746/services/bls_aggregation/blsagg.go#L263))
and the return condition is: 
- return whenever min threshold met
- if threshold hasnt been met after n seconds, return an err

AVSs that want low latency (at the cost of potential gas costs) need a more flexible API, in the form of:
- minimum signed stake threshold
- max # blocks of latency
- num blocks to expiry
with return condition being:
- as soon as min signed stake threshold is met, return if max # blocks of latency have expired, otherwise keep aggregating for that many blocks. (this way can keep aggregating to reduce gas costs, but with a max # blocks of latency SLA)
- if # blocks to expiry is met, return err

Mauro and Mariano, would this interface work for AlignedLayer?

TomasArrachea:
Closes #300 

Add BLS aggregation window-waiting-after-quorum feature. Once the aggregator reaches quorum, instead of sending the aggregated signature response, it will wait for a `windowDuration` time. During that window time, it will continue to receive and aggregate new signatures, allowing to reduce the gas cost of validating the aggregation on chain. When the window time finishes, the aggregation is sent.
In the case that task's expiry time is reached during that window, the aggregation response is sent.

To implement this feature, the aggregation loop was modified. There are 3 posible cases:
- receiving a new signature,
- expiration timer, 
- window timer.

Each time a new signature is received, it checks if quorum is reached. Once it's reached, the window timer fires. The loop continues until one of the two timers finish (window timer or task expiration).

This tests cases were added:
- signatures are processed during window after quorum
- if quorum has been reached and the task expires during window, the response is sent
- if window duration is zero, no signatures are aggregated after reaching quorum
- no signatures are aggregated after window

### Reviewer Checklist
- [ ] Code is well-documented

### DIFF
--- a/services/bls_aggregation/blsagg.go
+++ b/services/bls_aggregation/blsagg.go
@@ -81,18 +81,35 @@ type aggregatedOperators struct {
 // BlsAggregationService is the interface provided to avs aggregator code for doing bls aggregation
 // Currently its only implementation is the BlsAggregatorService, so see the comment there for more details
 type BlsAggregationService interface {
-	// InitializeNewTask should be called whenever a new task is created. ProcessNewSignature will return an error
-	// if the task it is trying to process has not been initialized yet.
-	// quorumNumbers and quorumThresholdPercentages set the requirements for this task to be considered complete, which
+	// InitializeNewTask creates a new task goroutine meant to process new signed task responses for that task
+	// (that are sent via ProcessNewSignature) and adds a channel to a.taskChans to send the signed task responses to it.
+	// The quorumNumbers and quorumThresholdPercentages set the requirements for this task to be considered complete, which
 	// happens when a particular TaskResponseDigest (received via the a.taskChans[taskIndex]) has been signed by signers
-	// whose stake
-	// in each of the listed quorums adds up to at least quorumThresholdPercentages[i] of the total stake in that quorum
+	// whose stake in each of the listed quorums adds up to at least quorumThresholdPercentages[i] of the total stake in
+	// that quorum
 	InitializeNewTask(
 		taskIndex types.TaskIndex,
 		taskCreatedBlock uint32,
 		quorumNumbers types.QuorumNums,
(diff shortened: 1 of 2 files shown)


services/bls_aggregation/blsagg.go
```diff
@@ -185,17 +205,43 @@ func (a *BlsAggregatorService) GetResponseChannel() <-chan BlsAggregationService
...
+	)
+}
+
+// InitializeNewTaskWithWindow creates a new task goroutine meant to process new signed task responses for that task
+// (that are sent via ProcessNewSignature) and adds a channel to a.taskChans to send the signed task responses to it.
+// The quorumNumbers and quorumThresholdPercentages set the requirements for this task to be considered complete, which
+// happens when a particular TaskResponseDigest (received via the a.taskChans[taskIndex]) has been signed by signers
+// whose stake in each of the listed quorums adds up to at least quorumThresholdPercentages[i] of the total stake in
+// that quorum.
+// Once the quorum is reached, the task is still open for a window of `windowDuration` time to receive more signatures,
```
afkbyte, line 238: add comment that if the windowDuration is longer than the taskexpiry, the task will still be sent to the aggregation response channel at taskExpiry

TomasArrachea: Done!

afkbyte (approved):

PR #392, opened 2024-11-07
Add exponential backoff retries to send transactions

TomasArrachea:
Fixes # .

### Reviewer Checklist
- [ ] Code is well-documented


chainio/txmgr/simple.go
```diff
@@ -91,6 +93,36 @@ func (m *SimpleTxManager) Send(
...
+		backoff.WithMultiplier(multiplier),
+		backoff.WithMaxElapsedTime(maxElapsedTime),
+	)
+
+	sendAndWait := func() (*types.Receipt, error) {
+		r, err := m.send(ctx, tx)
+		if err != nil {
+			return nil, err
+		}
+		m.logger.Error("failed to send transaction", err)
```
afkbyte: think this might be better at a warn level, and might be good to add context about the retry count as well

chainio/txmgr/simple.go
```diff
@@ -91,6 +93,36 @@ func (m *SimpleTxManager) Send(
...
+		m.logger.Debugf("waiting %f seconds for backoff", initialInterval.Seconds())
```
afkbyte: I think we need to use `NextBackoff` to get the next backoff interval

chainio/txmgr/simple.go
MegaRedHand: Done

chainio/txmgr/simple.go
MegaRedHand: `NextBackoff` increments the timeout interval, and I didn't find a way to get the next interval without incrementing.
I removed the log to avoid confusion.

afkbyte (approved):

PR #395, opened 2024-11-14
feat: signer abstraction

shrimalmadhur:
Fixes # .

### Reviewer Checklist
- [ ] Code is well-documented

### DIFF
--- a/crypto/bn254/cerberus/cerberus.go
+++ b/crypto/bn254/cerberus/cerberus.go
@@ -1,0 +1,16 @@
+package cerberus
+
+type Signer struct {
+}
+
+func New(signerUrl, pubKeyHex, password string) *Signer {
+	return &Signer{}
+}
+
+func (s *Signer) SignMessage(message [32]byte) ([]byte, error) {
+	return nil, nil
+}
+
+func (s *Signer) GetOperatorId() (string, error) {
+	return "", nil
+}
--- a/crypto/bn254/keystore/keystore.go
+++ b/crypto/bn254/keystore/keystore.go
@@ -1,0 +1,32 @@
+package keystore
+
+import (
+	"context"
+	"github.com/Layr-Labs/eigensdk-go/crypto/bls"
+	"github.com/Layr-Labs/eigensdk-go/types"
+)
+
+type Signer struct {
+	keyPair *bls.KeyPair
+}
+
+func New(path, password string) (*Signer, error) {
+	keyPair, err := bls.ReadPrivateKeyFromFile(path, password)
+	if err != nil {
+		return nil, err
+	}
+	return &Signer{
+		keyPair: keyPair,
+	}, nil
+}
+
+func (s *Signer) SignMessage(ctx context.Context, message [32]byte) ([]byte, error) {
+	signatureBytes := s.keyPair.SignMessage(message).Bytes()
+	var signature []byte
+	copy(signature[:], signatureBytes[:])
+	return signature, nil
+}
+
+func (s *Signer) GetOperatorId() (string, error) {
+	return types.OperatorIdFromG1Pubkey(s.keyPair.PubKey).LogValue().String(), nil
+}
--- a/crypto/bn254/signer.go
+++ b/crypto/bn254/signer.go
@@ -1,0 +1,9 @@
+package bn254
+
+import "context"
+
+type Signer interface {
+	SignMessage(ctx context.Context, message [32]byte) ([]byte, error)
+
+	GetOperatorId() (string, error)
+}


crypto/bn254/signer.go
```diff
@@ -0,0 +1,9 @@
+package bn254
+
+import "context"
+
+type Signer interface {
+	SignMessage(ctx context.Context, message [32]byte) ([]byte, error)
+
+	GetOperatorId() (string, error)
```
shrimalmadhur, line 8: should there be `context.Context` here? it's a pure operation - as in it doesn't depend on network in any case.

crypto/bn254/keystore/keystore.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/Layr-Labs/eigensdk-go/crypto/bls"
+	"github.com/Layr-Labs/eigensdk-go/types"
+)
+
+type Signer struct {
+	keyPair *bls.KeyPair
+}
+
+func New(path, password string) (*Signer, error) {
+	keyPair, err := bls.ReadPrivateKeyFromFile(path, password)
```
shrimalmadhur, line 14: this will need to be updated to handle the new format from https://github.com/Layr-Labs/bn254-keystore-go - will add that.

crypto/bn254/signer.go
samlaf: Then no. Unless you think some implementation in the future might need to make this call over the network, for eg calling a contract?

crypto/bn254/signer.go
```diff
@@ -0,0 +1,9 @@
+package bn254
```
samlaf, line 1: why does this all live in bn254? shouldnt signer interface be independent of signing scheme?

crypto/bn254/signer.go
samlaf: wait how does cerberus get operatorId? Doesn't it have to ask the remote signer over the network?

crypto/bn254/cerberus/cerberus.go
```diff
@@ -0,0 +1,16 @@
+package cerberus
+
+type Signer struct {
+}
+
+func New(signerUrl, pubKeyHex, password string) *Signer {
```
samlaf, line 6: what is password for here?

crypto/bn254/keystore/keystore.go
```diff
@@ -0,0 +1,32 @@
...
+	"context"
+	"github.com/Layr-Labs/eigensdk-go/crypto/bls"
+	"github.com/Layr-Labs/eigensdk-go/types"
+)
+
+type Signer struct {
+	keyPair *bls.KeyPair
+}
+
+func New(path, password string) (*Signer, error) {
```
samlaf, line 13: NewFromFile and NewFromHex?

crypto/bn254/keystore/keystore.go
```diff
@@ -0,0 +1,32 @@
...
+	}
+	return &Signer{
+		keyPair: keyPair,
+	}, nil
+}
+
+func (s *Signer) SignMessage(ctx context.Context, message [32]byte) ([]byte, error) {
+	signatureBytes := s.keyPair.SignMessage(message).Bytes()
+	var signature []byte
+	copy(signature[:], signatureBytes[:])
```
samlaf, line 26: why do you have to copy like this? signatureBytes is a fixed array [32]byte or something?

crypto/bn254/signer.go
shrimalmadhur: operatorId is just extracted out of public key. so that call just not need to go over the network since public key is provided by the caller when initializing the client.

crypto/bn254/signer.go
shrimalmadhur: hmm you are right actually - although I just want the signer to be for BLS signer - the ethereum signer will have different interfaces. We can have this signer support bn254 or bls12381. would that be okay?

crypto/bn254/cerberus/cerberus.go
shrimalmadhur: password to encrypt the keystore file stored in remote signer.

crypto/bn254/keystore/keystore.go
shrimalmadhur: this is always from File that's why keystore. maybe I can create another implementation for plaintext? but I don't want anyone to use plaintext 😅

crypto/bn254/keystore/keystore.go
shrimalmadhur: > signatureBytes is a fixed array [32]byte

Yep that's exactly why I need to copy.

PR #400, opened 2024-12-05
fix: anvil setup

shrimalmadhur:
fix anvil setup


contracts/script/parsers/EigenlayerContractsParser.sol
```diff
@@ -10,19 +10,21 @@ import {IStrategyManager, IStrategy} from "eigenlayer-contracts/src/contracts/in
...
+    struct EigenlayerContracts {
     ProxyAdmin eigenlayerProxyAdmin;
     PauserRegistry eigenlayerPauserReg;
     IStrategyManager strategyManager;
     IDelegationManager delegationManager;
-    ISlasher slasher;
     IAVSDirectory avsDirectory;
     IRewardsCoordinator rewardsCoordinator;
     StrategyBaseTVLLimits baseStrategyImplementation;
+    IAllocationManager allocationManager;
```
bdchatham, line 26: nit: should we add permission controller as well?

contracts/script/parsers/EigenlayerContractsParser.sol
```diff
@@ -65,12 +67,12 @@ contract EigenlayerContractsParser is ConfigsReadWriter {
                 ".addresses.avsDirectory"
             )
         );
-        ISlasher slasher = ISlasher(
-            stdJson.readAddress(
-                eigenlayerDeployedContracts,
-                ".addresses.slasher"
-            )
-        );
+//        ISlasher slasher = ISlasher(
```
bdchatham: nit: reminder to clean up if unintended.

bdchatham (approved):

contracts/script/parsers/EigenlayerContractsParser.sol
shrimalmadhur: I want to first fix it for existing test. but yes we can add that.

PR #401, opened 2024-12-06
feat: methods to set and read operator avs split

gpabst:
Adds convenience methods for reading and setting operator AVS split

### Reviewer Checklist
- [ ] Code is well-documented

### DIFF
--- a/chainio/clients/elcontracts/reader.go
+++ b/chainio/clients/elcontracts/reader.go
@@ -341,16 +341,16 @@ func (r *ChainReader) GetOperatorAVSSplit(
 	ctx context.Context,
 	operator gethcommon.Address,
 	avs gethcommon.Address,
-) (*uint16, error) {
+) (uint16, error) {
 	if r.rewardsCoordinator == nil {
-		return nil, errors.New("RewardsCoordinator contract not provided")
+		return 0, errors.New("RewardsCoordinator contract not provided")
 	}
 
 	split, err := r.rewardsCoordinator.GetOperatorAVSSplit(&bind.CallOpts{Context: ctx}, operator, avs)
 
 	if err != nil {
-		return nil, err
+		return 0, err
 	}
 
-	return &split, err
+	return split, nil
 }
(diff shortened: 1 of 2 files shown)


chainio/clients/elcontracts/reader.go
```diff
@@ -336,3 +336,21 @@ func (r *ChainReader) CheckClaim(
 
 	return r.rewardsCoordinator.CheckClaim(&bind.CallOpts{Context: ctx}, claim)
 }
+
+func (r *ChainReader) GetOperatorAVSSplit(
+	ctx context.Context,
+	operator gethcommon.Address,
+	avs gethcommon.Address,
+) (*uint16, error) {
```
shrimalmadhur: What's the point of returning a pointer here?

gpabst: So in can be nil in the case of no value since 0 is technically valid and I want to differentiate from that and no value set.

chainio/clients/elcontracts/reader.go
```diff
@@ -336,3 +336,21 @@ func (r *ChainReader) CheckClaim(
...
+		return nil, errors.New("RewardsCoordinator contract not provided")
+	}
+
+	split, err := r.rewardsCoordinator.GetOperatorAVSSplit(&bind.CallOpts{Context: ctx}, operator, avs)
+
+	if err != nil {
+		return nil, err
+	}
+
+	return &split, err
```
bdchatham: nit: err -> nil

chainio/clients/elcontracts/reader.go
bdchatham: I'm a little confused by this. What is the return type of `GetOperatorAVSSplit`?

bdchatham (dismissed):
Want to understand the reasoning on *uint16 and specifics on how that is going to work, but not an immediate blocker to releasing this.

chainio/clients/elcontracts/reader.go
shrimalmadhur: `split, err := r.rewardsCoordinator.GetOperatorAVSSplit(&bind.CallOpts{Context: ctx}, operator, avs)`
does this return an error if there's no split set?

shrimalmadhur: ok pointer does not makes sense here https://github.com/Layr-Labs/eigenlayer-contracts/blob/dev/src/contracts/core/RewardsCoordinator.sol#L635-L646

if nothing is set - it is returning default. so you won't be able to differentiate

bdchatham (approved):

PR #403, opened 2024-12-07
feat: bls signer abstraction

shrimalmadhur:
Fixes # .

### What Changed?
- Introducing a separate module for the signer. We are thinking of making eigensdk a multi module for better version management and usage experience. 
- Re-using the signer folder (which is deprecated) for new signer abstraction

### Reviewer Checklist
- [ ] Code is well-documented


ian-shim (commented):
lgtm, can we add some tests?

shrimalmadhur:
> lgtm, can we add some tests?

cool ya -  I will add those tests and then I will also integrate with eigenda

ian-shim:
when can we get back to this?

shrimalmadhur:
> when can we get back to this? 

Im going to prioritize this once we are back. There's some things where I need to update this to get both g1 and g2 pub keys as eigenda needs. Last week was busy with slashing release.

signer/bls/cerberus/cerberus.go
```diff
@@ -0,0 +1,129 @@
...
+			log.Fatalf("could not load tls cert: %s", err)
+		}
+		opts = append(opts, grpc.WithTransportCredentials(creds))
+	} else {
+		opts = append(opts, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	}
+
+	conn, err := grpc.NewClient(cfg.URL, opts...)
+	if err != nil {
+		log.Fatalf("did not connect: %v", err)
```
ian-shim: any reason why it's crashing instead of returning error?

signer/bls/cerberus/cerberus.go
```diff
@@ -0,0 +1,129 @@
...
+	return Signer{
+		signerClient: signerClient,
+		kmsClient:    kmsClient,
+		pubKeyHex:    cfg.PublicKeyHex,
+		password:     cfg.Password,
+	}, nil
+}
+
+func (s Signer) Sign(ctx context.Context, msg []byte) ([]byte, error) {
+	if len(msg) != 32 {
```
ian-shim, line 68: curious why the msg length has to be 32?

ian-shim (dismissed):

signer/bls/cerberus/cerberus.go
shrimalmadhur: yea because our underlying crypto library expect a 32 bytes array.  I think we might eventually want to change the underlying lib to support arbitrary bytes but right now I think we use 32 bytes everywhere so that's why making this validation.

signer/bls/cerberus/cerberus.go
shrimalmadhur: ah good catch, I will update.

shrimalmadhur: 2357268d8a189b99301d7d121135e6a84ea322ad

ian-shim (approved):

PR #410, opened 2024-12-17
add slashable shares

shrimalmadhur:
- add slashable shares per operators

### DIFF
--- a/chainio/clients/elcontracts/reader.go
+++ b/chainio/clients/elcontracts/reader.go
@@ -561,6 +561,44 @@ func (r *ChainReader) GetStrategiesForOperatorSet(
 	}
 }
 
+func (r *ChainReader) GetSlashableShares(
+	ctx context.Context,
+	operatorAddress gethcommon.Address,
+	operatorSet allocationmanager.OperatorSet,
+	strategies []gethcommon.Address,
+) (map[gethcommon.Address]*big.Int, error) {
+	if r.allocationManager == nil {
+		return nil, errors.New("AllocationManager contract not provided")
+	}
+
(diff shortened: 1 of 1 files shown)


chainio/clients/elcontracts/reader.go
```diff
@@ -561,60 +561,64 @@ func (r *ChainReader) GetStrategiesForOperatorSet(
...
 
-// GetDelegatedAndSlashableSharesForOperatorSetsBefore returns the strategies the operatorSets take into account, their
-// operators, and the minimum amount of shares that multiple operators delegated to them and slashable by the
+// GetSlashableSharesForOperatorSetsBefore returns the strategies the operatorSets take into account, their
+// operators, and the minimum amount of shares slashable by the
 // operatorSets before a given timestamp.
 // Timestamp must be in the future. Used to underestimate future slashable stake.
 // Not supported for M2 AVSs
-func (r *ChainReader) GetDelegatedAndSlashableSharesForOperatorSetsBefore(
+func (r *ChainReader) GetSlashableSharesForOperatorSetsBefore(
```
bdchatham, line 621: Minor nit: I always like when a method gives me the unit of the value compared to in the method name. In this case it would be `BlockTime`. Since Go doesn't support overloading, it seems like this would make sense idiomatically.

bdchatham (approved):

chainio/clients/elcontracts/reader.go
shrimalmadhur: oh shoot - i saw this after I merged. yeah I also don't like `Before` and a good name would be `GetSlashableSharesForOperatorSetsAtBlock`. will change it later

PR #415, opened 2024-12-23
[DO NOT MERGE] Update sdk to new contract layout with slashing feature branch

afkbyte:
### Reviewer Checklist
- [ ] Code is well-documented


whee
pablodeymo, line 1: I think this file should be removed from the PR.

chainio/clients/avsregistry/bindings.go
```diff
@@ -209,6 +211,11 @@ func NewBindingsFromConfig(
 		if err != nil {
 			return nil, utils.WrapError("Failed to get AvsDirectory address", err)
 		}
+		allocationManagerAddr, err = contractServiceManager.AllocationManager(&bind.CallOpts{})
+		if err != nil {
+			return nil, utils.WrapError("Failed to get AllocationManager address", err)
+		}
+
 	}
```
pablodeymo, line 220: Here, we could add the same idea of the check as in elcontracts:
```
if isZeroAddress(cfg.AllocationManagerAddress) {
...
```

if we don't get the address of the allocation Manager, we should ignore it while creating the return of the function.

pablodeymo:
It should be good to make this compatible with previous version of AVSs contract code.
To do so, we can ignore missing allocation manager addresses and allocation manager function in serviceManager.

shrimalmadhur:
@afkbyte we have been working off this branch for core contracts and other slashing changes. I think we are duplicating some work here https://github.com/Layr-Labs/eigensdk-go/pull/342

shrimalmadhur:
> @afkbyte we have been working off this branch for core contracts and other slashing changes. I think we are duplicating some work here https://github.com/Layr-Labs/eigensdk-go/pull/342

The only thing my branch lacks is new middleware bindings and respective tests. I think we should fork out of my branch and then add those. Wdyt?

haardikk21:
Hey folks - since the holesky contracts have been upgraded, this PR is blocking AVS testing on testnet as latest release of eigensdk-go attempts to call a `slasher()` view function on the delegation manager which no longer exists

It's the holidays so no pressure at all, but just out of curiosity do you have an ETA for this PR (or #342) to be merged?

afkbyte:
closing in favour of #342

PR #416, opened 2024-12-27
test and fix `ChainWriter.RegisterForOperatorSets` function

pablodeymo:
### What Changed?
This PR introduces the following changes:
* Adds a test for the `ChainWriter` `RegisterForOperatorSets` and `deregisterFromOperatorSets` functions.
* Fixes the `ChainWriter.RegisterForOperatorSets`. Adds some missing parameters (socket and public key registration params) to make it work.

Note that we introduced a new argument, `registryCoordinatorAddress`, to the `RegisterForOperatorSets` function. This addition was necessary to retrieve the missing registration parameters. 
However we are uncertain about this addition as it changes the function interface. Please let us know what you think about it.

The main two options we have for the interface are:
- the current solution (handling the bytes encoding ourselves), which makes the function easier to use;
- having the function receive arbitrary bytes, which makes the function more flexible (we could also disponibilize an encoding utility)

### Reviewer Checklist
- [ ] Code is well-documented


afkbyte (approved):
lgtm!

PR #426, opened 2025-01-07
chore: remove duplicated function

MegaRedHand:
`ForceDeregisterFromOperatorSets` does the same as `DeregisterFromOperatorSets`, so we can simply remove it.


chainio/clients/elcontracts/writer.go
```diff
@@ -466,48 +466,6 @@ func (w *ChainWriter) ProcessClaims(
...
-	receipt, err := w.txMgr.Send(ctx, tx, waitForReceipt)
-	if err != nil {
-		return nil, utils.WrapError("failed to send tx", err)
-	}
-
-	return receipt, nil
-}
-
-// Modifies the proportions of slashable stake allocated to an operator set
-// from a list of strategies, for the given `operatorAddress`.
```
damiramirez: I think you are also deleting the comments from the next method.

MegaRedHand: Good catch!

pablodeymo: I agree with Damián

pablodeymo (approved):

PR #442, opened 2025-01-16
fix: change literal string values for constants or predefined values

maximopalopoli:
### What Changed?
Changes literal values of the assigned `RewardsCoordinator` for the field of the `ContractAddresses` util, a better way to use this "constant" values. Also, replaces literal appearances of Anvil first `Addresses`/`privateKeys` for the constants from `testutils` package.

### Reviewer Checklist
- [ ] Code is well-documented


testutils/anvil.go
```diff
@@ -20,7 +20,7 @@ import (
 )
 
 const (
-	ANVIL_FIRST_ADDRESS      = "f39Fd6e51aad88F6F4ce6aB8827279cffFb92266"
+	ANVIL_FIRST_ADDRESS      = "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266"
```
MegaRedHand, line 23: Why the leading `0x`?

maximopalopoli: It was added because a test failed in which the conversion from address to string added a 0x in front ([this one](https://github.com/Layr-Labs/eigensdk-go/blob/fda56ad16fad8d29f63bf376dd9d7cf6f32f012e/chainio/clients/elcontracts/reader_test.go#L141)), and I preferred to add it to the constant rather than adding it when doing the assert.

MegaRedHand (approved):
LGTM

PR #446, opened 2025-01-17
test: add test cases where contract returns an error

maximopalopoli:
### What Changed?
The initial idea was to test the cases in which the binding response returned an error. I was able to do it only in two cases, in the others, since the chain reader is mainly a getter, it was difficult to get it to return an error. Makes test coverage of `chainio/clients/elcontracts/reader.go` increase in 1%.

Also in this PR some verifications on binding result are removed, since they only added a little description on the returned error, but the error was extended anyway. The verification removed were those that didn't add context to the error, and those that added context, like those calling permissionController contract were not removed.

### Reviewer Checklist
- [ ] Code is well-documented


damiramirez:
LGTM

pablodeymo (approved):

PR #464, opened 2025-01-23
feat: make SDK compatible with mainnet contracts

MegaRedHand:
### What Changed?

This PR makes the `AllocationManager` initialization optional in the bindings builder. This should make the SDK compatible with the current version of the contracts in mainnet.

### Reviewer Checklist
- [ ] Code is well-documented


chainio/clients/avsregistry/bindings.go
```diff
@@ -172,6 +176,15 @@ func NewBindingsFromConfig(
 	}, nil
 }
 
+// Makes a contract call to retrieve an address, but returning a zero address if the call reverts.
+func getAddressOrZero(getAddress func(*bind.CallOpts) (gethcommon.Address, error)) (gethcommon.Address, error) {
+	address, err := getAddress(&bind.CallOpts{})
+	if errors.Is(err, gethvm.ErrExecutionReverted) {
```
MegaRedHand: This is a `jsonError`, and even though it results in the same error message, we can't match on it 😭

pablodeymo (approved):

PR #465, opened 2025-01-23
refactor: `elcontracts/reader` interface

damiramirez:
### What Changed?

This PR aims to improve the `elcontracts/reader` interface by implementing a request-response pattern. For the exposed functions, we create structs that represent the request (specific fields + `BlockNumber`) and the response.

Because the function signatures change, we also refactor the tests to align with the new pattern.

Below is a simple example of the new request-response pattern:

Example:
```go
func(r *ChainReader) Foo(
	ctx context.Context,
	request FooRequest,
) (FooResponse, error) {
	...
}

type FooRequest struct {
	BlockNumber *big.Int,
        ...
}
```

This is a work in progress and changes may occur.

### Reviewer Checklist
- [ ] Code is well-documented

### DIFF
--- a/chainio/clients/avsregistry/writer.go
+++ b/chainio/clients/avsregistry/writer.go
@@ -28,11 +28,9 @@ import (
 type eLReader interface {
 	CalculateOperatorAVSRegistrationDigestHash(
 		ctx context.Context,
-		operatorAddr gethcommon.Address,
-		serviceManagerAddr gethcommon.Address,
-		operatorToAvsRegistrationSigSalt [32]byte,
-		operatorToAvsRegistrationSigExpiry *big.Int,
-	) ([32]byte, error)
+		blockNumber *big.Int,
+		request elcontracts.CalculateOperatorAVSRegistrationDigestHashRequest,
+	) (elcontracts.CalculateOperatorAVSRegistrationDigestHashResponse, error)
 }
 
 type ChainWriter struct {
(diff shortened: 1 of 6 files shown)


chainio/clients/avsregistry/writer.go
```diff
@@ -161,17 +158,21 @@ func (w *ChainWriter) RegisterOperatorInQuorumWithAVSRegistryCoordinator(
 	}
 
 	// params to register operator in delegation manager's operator-avs mapping
-	msgToSign, err := w.elReader.CalculateOperatorAVSRegistrationDigestHash(
+	// TODO: Review this function after finishing the refactor of the ChainReader
```
maximopalopoli: Question, should this line be left in? Or is the chainReader refactoring not finished yet? The same in the other call to `elcontracts.CalculateOperatorAVSRegistrationDigestHashRequest`

maximopalopoli: Same in the TODO of elcontracts's ChainWriter related to this

maximopalopoli (commented):
LGTM, just leave you a comment:

chainio/clients/avsregistry/writer.go
damiramirez: I added that comment because the `avsregistry` and `elcontracts/writer` refactors are being handled in separate PRs. It might be unnecessary to keep this comment since conflicts are expected during merging and will need to be handled regardless. So I will remove the comments!

damiramirez: Done [d110c54](https://github.com/Layr-Labs/eigensdk-go/pull/465/commits/d110c5466aebd8efda9b6a24d1d59fce8881db69)!

PR #477, opened 2025-01-27
refactor: common error interface for elcontracts chainReader

maximopalopoli:
### What Changed?

The idea is to create a common interface to all SDK errors. The structure of this common interface is currently:
``` Go
type Error struct {
	code        int
	message     string
	description string
	cause       error
}
```
The code is associated with the message, which could change in the future, such that one of both must be kept and the other must be replaced or transformed.

One use example could be the following:

``` Go
func (r *ChainReader) IsOperatorRegistered(
	ctx context.Context,
	operator types.Operator,
) (bool, error) {
	if r.delegationManager == nil {
		wrappedError := MissingContractError("DelegationManager")
		return false, wrappedError
	}
...
}

func MissingContractError(contractName string) Error {
	errDescription := fmt.Sprintf("%s contract not provided", contractName)
	return Error{1, "Missing needed contract", errDescription, nil}
}

```
So, in this use example the error message would be the following:

`"Missing needed contract(1) - DelegationManager contract not provided"`

### Reviewer Checklist
- [ ] Code is well-documented


chainio/clients/elcontracts/error.go
```diff
@@ -0,0 +1,48 @@
+package elcontracts
+
+import "fmt"
+
+type Error struct {
+	code        int
```
damiramirez, line 6: What do you think about creating a type of "enum" for the errors codes? Maybe it would be more expressive?

I think that we can do something like this:

```go
type ErrorCode int

const (
    ErrCodeBindingError ErrorCode = 0
    ErrCodeMissingContract
    ErrCodeNestedError
)
```

Do you think this is a good option? If not, we can use the current implementation.

chainio/clients/elcontracts/error.go
```diff
@@ -0,0 +1,48 @@
...
+	message     string
+	description string
+	cause       error
+	// metadata    map[string]interface{}
```
damiramirez: Do we need this?

chainio/clients/elcontracts/error.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+func (e Error) Error() string {
+	if e.cause != nil {
+		return fmt.Sprintf("%s(%d) - %s: %s", e.message, e.code, e.description, e.cause.Error())
+	} else {
+		return fmt.Sprintf("%s(%d) - %s", e.message, e.code, e.description)
+	}
+}
```
damiramirez:
```suggestion
func (e Error) Error() string {
	if e.cause != nil {
		return fmt.Sprintf("%s(%d) - %s: %s", e.message, e.code, e.description, e.cause.Error())
	}
	return fmt.Sprintf("%s(%d) - %s", e.message, e.code, e.description)
}
```

damiramirez:
LGTM! Just left some comments

chainio/clients/elcontracts/error.go
maximopalopoli: No, I left it just in case but could not find any use. Removed in [6978c58](https://github.com/Layr-Labs/eigensdk-go/pull/477/commits/6978c58de04fcbaf3586307da739f3583bc72552)

chainio/clients/elcontracts/error.go
maximopalopoli: Yes, I thought about leaving a comment or something in the code that clarifies what type of error each ID is equivalent to, but I would prefer not to do so until I clearly define what error each one is equivalent to (it could change)

chainio/clients/elcontracts/error.go
```diff
@@ -0,0 +1,50 @@
...
+		return fmt.Sprintf("%s(%d) - %s: %s", e.message, e.code, e.description, e.cause.Error())
+	}
+	return fmt.Sprintf("%s(%d) - %s", e.message, e.code, e.description)
+}
+
+func (e Error) Unwrap() error {
+	return e.cause
+}
+
+func CreateErrorForMissingContract(contractName string) Error {
```
MegaRedHand: Let's rename these functions to `MissingContractError`, `BindingError`, `XError`, etc.

chainio/clients/elcontracts/reader.go
```diff
@@ -40,7 +38,7 @@ type ChainReader struct {
 	ethClient            eth.HttpBackend
 }
 
-var errLegacyAVSsNotSupported = errors.New("method not supported for legacy AVSs")
+var errLegacyAVSsNotSupported = Error{3, "Other errors", "Method not supported for legacy AVSs", nil}
```
MegaRedHand: Let's introduce another function for this kind of "Other errors", although we could probably find a category to move it to.

chainio/clients/elcontracts/error.go
```diff
@@ -0,0 +1,50 @@
...
+	return Error{1, "Missing needed contract", errDescription, nil}
+}
+
+func CreateForBindingError(bindingName string, errorCause error) Error {
+	errDescription := fmt.Sprintf("Error happened while calling %s", bindingName)
+	return Error{
+		0,
+		"Binding error",
+		errDescription,
+		errorCause,
```
MegaRedHand: Let's use the full syntax here:

```suggestion
		code: 0,
		message: "Binding error",
		description: errDescription,
		cause: errorCause,
```

chainio/clients/elcontracts/reader.go
```diff
@@ -140,9 +166,11 @@ func (r *ChainReader) GetOperatorDetails(
 	)
 	// This call should not fail since it's a getter
 	if err != nil {
-		return types.Operator{}, err
+		wrappedError := CreateForBindingError("delegationManager.DelegationApprover", err)
```
MegaRedHand: Let's make it "DelegationManager.delegationApprover" following the solidity casing.

chainio/clients/elcontracts/reader_test.go
```diff
@@ -832,10 +847,16 @@ func TestInvalidConfig(t *testing.T) {
...
+		assert.Equal(t, err.Error(), elcontracts.CommonErrorMissingContract("DelegationManager"))
 
 		// GetStrategyAndUnderlyingToken needs a correct StrategyAddress
 		_, _, err = chainReader.GetStrategyAndUnderlyingToken(context.Background(), strategyAddr)
 		require.Error(t, err)
+		assert.Equal(
+			t,
+			err.Error(),
+			"Binding error(0) - Error happened while calling token contract: no contract code at given address",
+		)
```
MegaRedHand: We can use [`ErrorContains`](https://pkg.go.dev/github.com/stretchr/testify/require#ErrorContains) here.

chainio/clients/elcontracts/reader_test.go
```diff
@@ -1224,12 +1267,15 @@ func TestOperatorSetsWithWrongInput(t *testing.T) {
...
 		require.Error(t, err)
+		assert.Equal(t, err.Error(), "Other errors(3) - Method not supported for legacy AVSs")
 
 		_, err = chainReader.GetNumOperatorsForOperatorSet(ctx, operatorSet)
 		require.Error(t, err)
+		assert.Equal(t, err.Error(), "Other errors(3) - Method not supported for legacy AVSs")
 
 		_, err = chainReader.GetStrategiesForOperatorSet(ctx, operatorSet)
 		require.Error(t, err)
+		assert.Equal(t, err.Error(), "Other errors(3) - Method not supported for legacy AVSs")
```
MegaRedHand: We should expose this as a variable.

chainio/clients/elcontracts/error.go
maximopalopoli: Done in [Rename custom error builders and its occurrences](https://github.com/Layr-Labs/eigensdk-go/pull/477/commits/6ce8886b03085f34cf2710edca3bdcc3266743d7)

chainio/clients/elcontracts/reader.go
maximopalopoli: Yes, I thought about renaming some of them to "logic errors", but preferred to keep "others" for now.

chainio/clients/elcontracts/error.go
maximopalopoli: Done in [Use full syntax when creating errors](https://github.com/Layr-Labs/eigensdk-go/pull/477/commits/b56b255007b0fbb1702b04f4b4af83d6fa1199af)

chainio/clients/elcontracts/reader.go
maximopalopoli: Done in [Rename contracts names at errors following the solidity casing](https://github.com/Layr-Labs/eigensdk-go/pull/477/commits/7e6a276cdc4ed0af3e0139cec35ce7959c5ed75d)

chainio/clients/elcontracts/reader_test.go
maximopalopoli: Done in [Use ErrorContains instead of comparing long literals](https://github.com/Layr-Labs/eigensdk-go/pull/477/commits/3fda64757681645dd37a0a8ea19dd91d80bfcc0f)

chainio/clients/elcontracts/reader.go
maximopalopoli: Done in [Create and use OtherError msg wrapper](https://github.com/Layr-Labs/eigensdk-go/pull/477/commits/15ed4db5c24e5a59965cb976e4837d3dc2e30bf8)

chainio/clients/elcontracts/reader_test.go
maximopalopoli: Done in [Create only once legacyAVSsNotSupportedErrorMsg string](https://github.com/Layr-Labs/eigensdk-go/pull/477/commits/2f4befa8fd386d1ce8c7ca8683221c19a7180195)

PR #482, opened 2025-01-28
refactor: common error interface for elcontracts chainWriter

maximopalopoli:
### What Changed?
This PR can be seen as a continuation of the PR #477, since expands the usage of the common error interface to the `elcontracts` chainWriter.

### Reviewer Checklist
- [ ] Code is well-documented


chainio/clients/elcontracts/error.go
```diff
@@ -0,0 +1,66 @@
+package elcontracts
+
+import "fmt"
+
+type Error struct {
+	code        int
+	message     string
+	description string
+	cause       error
```
MegaRedHand, line 9: We should add an [`Unwrap` method](https://pkg.go.dev/errors)

maximopalopoli: Done in [Add Unwrap method to get the underlying error](https://github.com/Layr-Labs/eigensdk-go/pull/482/commits/2b5779411afd3875144cc8d3e7f5b676dc44724b)

PR #458, opened 2025-01-22
test: `avsregistry` add new tests cases for reader methods

damiramirez:
### What Changed?

- Added a helper function to create a `chainReader` for `avsregistry`. This enables testing scenarios like using an empty configuration.
- Added missing tests identified in the coverage report.
- Introduced additional tests to ensure that `chainReader` methods properly handle errors when an invalid configuration is provided. 

### Reviewer Checklist
- [ ] Code is well-documented
- [ ] Code adheres to Go [naming conventions](https://go.dev/doc/effective_go#names)
- [ ] Code deprecates any old functionality before removing it

PR #486, opened 2025-01-29
refactor: common error interface for avsRegistry chainReader

maximopalopoli:
### What Changed?
This PR can be seen as a continuation of the PRs #477 and #482, since expands the usage of the common custom error to the `avsRegistry` chainReader.

The order of the error codes, the names of the errors themselves, or even the `Error` struct attributes may change in future PRs.

Note: this PR updates tests merged from #458 and should be merged after it.

### Reviewer Checklist
- [ ] Code is well-documented


maximopalopoli:
Previous to merge we should make the fixes suggested in #477 and also resolve conflicts.

PR #489, opened 2025-01-30
refactor: common Error struct used in avsRegistry chainWriter and chainSubscriber

maximopalopoli:
### What Changed?
This PR can be seen as a continuation of the PRs #477, #482 and #486, since expands the usage of the common custom error to the `avsRegistry` chainWriter and chainSubscriber.

The order of the error codes, the names of the errors themselves, or even the `Error` struct attributes may change in future PRs.

### Reviewer Checklist
- [ ] Code is well-documented


maximopalopoli:
Previous to merge we should make the fixes suggested in #477 and also resolve conflicts.

PR #508, opened 2025-02-04
feat: add clear deallocation queue function

maximopalopoli:
### What Changed?
Adds clearDeallocationQueue function to SDK with its corresponding tests.

### Reviewer Checklist
- [ ] Code is well-documented

### DIFF
--- a/chainio/clients/elcontracts/writer.go
+++ b/chainio/clients/elcontracts/writer.go
@@ -508,6 +508,35 @@ func (w *ChainWriter) ModifyAllocations(
 	return receipt, nil
 }
 
+func (w *ChainWriter) ClearDeallocationQueue(
+	ctx context.Context,
+	operatorAddress gethcommon.Address,
+	strategies []gethcommon.Address,
+	numsToClear []uint16,
+	waitForReceipt bool,
+) (*gethtypes.Receipt, error) {
+	if w.allocationManager == nil {
+		return nil, errors.New("AllocationManager contract not provided")
+	}
+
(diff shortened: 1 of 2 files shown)


chainio/clients/elcontracts/writer.go
```diff
@@ -508,6 +508,38 @@ func (w *ChainWriter) ModifyAllocations(
...
 }
 
+// Receives an operator address, and a list of strategies and numsToClear (number of elements to clear from queue),
+// and clears the operators deallocation queue in numbers to clear for the given strategies, by completing the
+// pending deallocations if their effect timestamps have passed.
+func (w *ChainWriter) ClearDeallocationQueue(
+	ctx context.Context,
+	operatorAddress gethcommon.Address,
+	strategies []gethcommon.Address,
+	numsToClear []uint16,
```
MegaRedHand, line 519: It would be good to merge these two args since the arrays must be the same length.

maximopalopoli: I thought about that change, but I didn't end up making it because the binding function receives them separately, not together. What I could do is make a struct request for the function as the roadmap suggests.

maximopalopoli: Perhaps it could be specified more clearly that the length of both arrays should be the same.

MegaRedHand: Let's document it and leave it for a later refactor

MegaRedHand (approved):
LGTM

PR #567, opened 2025-02-13
feat: Rewards coordinator storage part2

Sidu28:
Fixes # .

### Reviewer Checklist
- [ ] Code is well-documented


contracts/bindings/RewardsCoordinator/binding.go
MegaRedHand, line 1: Makefile changes to produce these bindings are missing.

chainio/clients/elcontracts/reader.go
```diff
@@ -515,6 +515,92 @@ func (r *ChainReader) GetClaimerFor(
...
+	avs gethcommon.Address,
+	hash [32]byte,
+) (bool, error) {
+	if r.rewardsCoordinator == nil {
+		return false, errors.New("RewardsCoordinator contract not provided")
+	}
+	return r.rewardsCoordinator.IsRewardsSubmissionForAllEarnersHash(&bind.CallOpts{Context: ctx}, avs, hash)
+}
+
+// Returns whether a hash is a valid operator set performance rewards submission hash for a given avs
```
MegaRedHand, line 576: This doc is the same as `GetIsOperatorDirectedOperatorSetRewardsSubmissionHash`. What's the difference here? I checked the contracts and the documentation is the same for both fields, so let's keep it as-is. I leave this comment here for posterity 😞

contracts/bindings/RewardsCoordinator/binding.go
Sidu28: sorry did not push my last commit, fixed

MegaRedHand (commented):
Can you regenerate the bindings? Since `IRewardsCoordinator` was removed from the `Makefile`, the bindings should be removed too.

chainio/clients/elcontracts/reader.go
Sidu28: Haha thats what the comments are in the smart contract:
https://github.com/Layr-Labs/eigenlayer-contracts/blob/499fd45503de7a5fe24b76d7fe1231dea18ac12d/src/contracts/core/RewardsCoordinatorStorage.sol#L120

https://github.com/Layr-Labs/eigenlayer-contracts/blob/499fd45503de7a5fe24b76d7fe1231dea18ac12d/src/contracts/core/RewardsCoordinatorStorage.sol#L132

Sidu28: Yeah I just copied it over from there

Sidu28:
> Can you regenerate the bindings? Since `IRewardsCoordinator` was removed from the `Makefile`, the bindings should be removed too.

@MegaRedHand removed unused bindings

MegaRedHand (approved):

PR #584, opened 2025-02-25
fix: use getAllocatedStake to get stake regardless of slashability status

gpabst:
We currently use the `getMinimumSlashableStake` function. This method has been updated recently to not return any stake if the operator is not slashable for the operatorSet. Therefore we need to update this to use the `getAllocatedStake` function to get the stake allocated by an operator to an operatorSet regardless of the operator's slashability status.

### Reviewer Checklist
- [ ] Code is well-documented


MegaRedHand:
Hey @gpabst! The `GetSlashableShares` function is intended to return the shares that are slashable for the operator. To get the allocated stake for the operator, you can use `GetAllocatedStake` instead

PR #598, opened 2025-03-09
chore: fix error comparison with `errors.Is`

0xkazak:
### What Changed?

I noticed that comparing errors using `err.Error() == maxByteErr.Error()` could be unreliable, as the error message text might change in future Go versions. Replaced the comparison with `errors.Is` to correctly check the error type, ensuring stability across Go updates.

### Reviewer Checklist
- [ ] Code is well-documented


pablodeymo:
It does not compile

PR #634, opened 2025-04-14
fix: improve v2 interfaces

maximopalopoli:
### What Changed?
This PR makes various changes in the new generic implementations of v2.

The changes are:
- Change those interface parameters of type any to types.Log, removing the any to Log parsing.
- Move the bindings of the example to a common folder, to avoid repeating them.

### Reviewer Checklist
- [ ] Code is well-documented


pablodeymo (approved):

aggregator/aggregator.go
```diff
@@ -18,23 +18,24 @@ import (
...
+
 	// aggregation related fields
 	blsAggregationService blsagg.BlsAggregationService
 	taskProcessor         TaskProcessor
 	newTaskCreatedLogs    chan types.Log
 }
 
 // NewAggregator creates a new Aggregator with the provided config.
-func NewAggregator(c AggregatorConfig, taskProcessor TaskProcessor, eventHash common.Hash) (*Aggregator, error) {
+func NewAggregator[ResponseType any](c AggregatorConfig, taskProcessor TaskProcessor, eventHash common.Hash) (*Aggregator[ResponseType], error) {
```
MegaRedHand, line 38: Doesn't this mean the implementor needs to specify the ResponseType? We should check that the generic is resolving correctly.

operator/operator.go
```diff
@@ -43,17 +43,22 @@ type OperatorTaskProcessor interface {
...
-	aggregatorRpcClient AggregatorRpcClienter
+	aggregatorRpcClient AggregatorRpcClienter[ResponseType]
 	EthWsUrl            string
 	blsKeypair          *bls.KeyPair
 	taskProcessor       OperatorTaskProcessor
 	newTaskCreatedLogs  chan types.Log
 }
 
-func NewOperatorFromConfig(c OperatorConfig, eventHash common.Hash, taskProcessor OperatorTaskProcessor, logger logging.Logger) (*Operator, error) {
+func NewOperatorFromConfig[ResponseType any](
```
MegaRedHand, line 56: Same here. We should find some way to auto-derive this for the user.

aggregator/aggregator.go
maximopalopoli: Yes, we should use a generic response type instead of requiring one, done on #640

operator/operator.go
maximopalopoli: The same as the previous comment, fixed in #640, so the user does not have to specify the type used, which will be inferred from the task processor.

pablodeymo (approved):

MegaRedHand (approved):

PR #635, opened 2025-04-14
refactor: `TaskGenerator` rewrite

MegaRedHand:
Fixes # .

### What Changed?

This PR rewrites the `TaskGenerator` to reduce the amount of code an AVS would need to write.

### Reviewer Checklist
- [ ] Code is well-documented


task-generator/task_generator.go
```diff
@@ -2,26 +2,38 @@ package taskgenerator
...
 	"github.com/Layr-Labs/eigensdk-go/logging"
+	"github.com/Layr-Labs/eigensdk-go/utils"
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	gethtypes "github.com/ethereum/go-ethereum/core/types"
 )
 
-type TaskGenerator struct {
+// Type is generic over the task input type
+type TaskManager[Input any] interface {
+	CreateNewTask(opts *bind.TransactOpts, input Input, quorumThresholdPercentage uint32, quorumNumbers []byte) (*gethtypes.Transaction, error)
```
MegaRedHand, line 16: I don't like that this requires the implementor to return the `Transaction` because it limits the kind of things they could do inside of this. However, one could hack around this restriction by simply returning an empty transaction and nil error, and using a custom `TxManager` that ignores all transactions.

MegaRedHand:
I left some TODOs in the code, but they can be addressed later. The main idea is whole.

task-generator/task_generator.go
```diff
@@ -30,32 +42,63 @@ func (taskGen *TaskGenerator) Start(ctx context.Context) error {
...
 
-type TaskGeneratorLogic interface {
-	SendNewTask(taskNumber int64) error
+func (taskGen *TaskGenerator[Input]) CreateNewTask(
+	ctx context.Context,
+	input Input,
+) error {
+	// TODO: make configurable
+	var quorumThresholdPercentage uint8 = 100
+	var quorumNumbers []uint8 = []uint8{0}
```
maximopalopoli: Can these be received by parameter in the NewTaskGenerator function?

MegaRedHand: Yes. That's the idea. I just left it out of this PR to reduce noise.

MegaRedHand: We should also change `NewTaskGenerator` to be:

```go
func NewTaskGenerator[Input any](logger logging.Logger, taskManager TaskManager[Input], tgConfig TaskGeneratorConfig) (*TaskGenerator[Input], error) {
```

Then `TaskGeneratorConfig` can be used as a kind of "builder", where users can specify only the fields they want to and others will default to sane values.

MegaRedHand:
I added a configuration struct with the quorum parameters, logger, and time between tasks (which I changed to be of type `time.Duration`).

I also tried doing a refactor to split the interface in two, which I found too complex for now (so I reverted it), but we might want to revisit it later: https://github.com/Layr-Labs/eigensdk-go/pull/635/commits/adc5dd25b814415825639e94e1c148e05a1a2b5e

task-generator/task_generator.go
pablodeymo: It's a good catch. However, I think we can leave this change for a future change.

pablodeymo (approved):

PR #648, opened 2025-04-24
fix: change aggregator building

maximopalopoli:
### What Changed?
This PR makes minor changes to the aggregator building. Those changes are:
- Removed `taskManagerABI` and `eventHash` parameters from the `NewAggregator` function. The first was moved to the AggregatorConfig, and the second is now calculated with the Abi inside the function.
- Now the `NewAggregator` function uses the AvsReader created in the `sdkclients.BuildAll()` function instead of creating another.
- Removed the TaskResponses map since it was unused
- Moved the seconds unit outside of the blockTimeSeconds constant.

### Reviewer Checklist
- [ ] Code is well-documented


aggregator/aggregator.go
```diff
@@ -29,7 +28,7 @@ const (
 	// number of blocks after which a task is considered expired this hardcoded here because it's also
 	//  hardcoded in the contracts, but should ideally be fetched from the contracts
 	taskChallengeWindowBlock = 100
-	blockTimeSeconds         = 12 * time.Second
+	blockTimeSeconds         = 12
```
MegaRedHand: Why undo this change?

maximopalopoli: Because there is a lint alert that says that the constant should not include duration:
`var blockTimeSeconds is of type time.Duration; don't use unit-specific suffix "Seconds" (ST1011)go-staticcheck`
I don't think the error applies, but it's annoying to have the linter.

MegaRedHand: Can't we just ignore that warning? In any case, the warning is not due to using `time.Second` in the value, but because the name includes "Seconds"

maximopalopoli: Yes, done in [eed650d](https://github.com/Layr-Labs/eigensdk-go/pull/648/commits/eed650da4f3b81230846ca75c7349249b6f80d33)

MegaRedHand (approved):

PR #677, opened 2025-05-13
fix(v2): challenger response validation interpretation

ricomateo:
### What Changed?
This PR changes the challenger response validation function to return whether the response is correct, instead of returning whether the response is wrong, to match the Rust implementation.

### Reviewer Checklist
- [ ] Code is well-documented

### DIFF
--- a/challenger/challenger-processor/challenger_task_processor.go
+++ b/challenger/challenger-processor/challenger_task_processor.go
@@ -47,13 +47,13 @@ func (icp IndexingChallengerProcessor[Input, Output]) ProcessTaskResponded(taskI
 		return fmt.Errorf("could not find the task for the received task index")
 	}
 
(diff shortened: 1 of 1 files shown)


challenger/challenger-processor/challenger_task_processor.go
```diff
@@ -68,7 +68,7 @@ func (icp IndexingChallengerProcessor[Input, Output]) ProcessTaskResponded(taskI
 
 // Takes a ResponseCalculator and an Equal function.
 // Returns a function that receives a task input and output, calculates the expected
-// output using the ResponseCalculator and returns whether is different to the given output,
+// output using the ResponseCalculator and returns whether is equal to the given output,
```
MegaRedHand:
```suggestion
// output using the ResponseCalculator and returns whether it is equal to the given output,
```

MegaRedHand (approved):

PR #715, opened 2025-05-30
fix: remove duplicate config option

MegaRedHand:
### What Changed?

The Challenger config already had an `EthHttpUrl` field, and declaring a duplicate like we did seems to cause the value for that to be missing.

### Reviewer Checklist
- [ ] Code is well-documented


ricomateo (approved):
LGTM

PR #717, opened 2025-06-02
fix(v2): improve asserts on avs integration test

maximopalopoli:
### What Changed?
This PR makes the following changes to the AVS integration test:
- Make the task spammer generate only three tasks
- Remove the operator failure
- Check that three tasks were submitted
- Check that the three task responses are not zero
- Rename the test file to add tests for the other example AVSs in the future.

### Reviewer Checklist
- [ ] Code is well-documented


integration-tests/incredible_squaring_test.go
```diff
@@ -78,15 +78,25 @@ func TestIncredibleSquaring(t *testing.T) {
 	case err := <-opErrC:
 		t.Fatal("Operator error:", err)
 	case err := <-tsErrC:
-		t.Fatal("Task Spammer error:", err)
+		if err == nil {
+			cancel()
+		} else {
+			t.Fatal("Task Spammer error:", err)
+		}
```
damiramirez, line 86: Why did we change this? Should we apply the same change above?

maximopalopoli: Explained in a comment in [213d0d5](https://github.com/Layr-Labs/eigensdk-go/pull/717/commits/213d0d58f943126fcf802223398dafc46bad0448)

MegaRedHand (approved):
Great!

PR #728, opened 2025-06-04
feat(v2): Check that computed and on-chain operator IDs are equal

maximopalopoli:
### What Changed?
This PR adds a check that the calculated and on-chain operator IDs are equal.

### Reviewer Checklist
- [ ] Code is well-documented

### DIFF
--- a/operator/operator.go
+++ b/operator/operator.go
@@ -125,9 +125,9 @@ func NewOperator[Input any, Output any](
 		return nil, err
 	}
 
-	calculatedOperatorId := sdktypes.OperatorIdFromKeyPair(blsKeyPair)
-	if operatorId != calculatedOperatorId {
-		return nil, fmt.Errorf("the operator Id calculated from BLS public key does not match the on chain operator Id")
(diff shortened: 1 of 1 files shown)


MegaRedHand (approved):

operator/operator.go
```diff
@@ -125,6 +125,11 @@ func NewOperator[Input any, Output any](
 		return nil, err
 	}
 
+	calculatedOperatorId := sdktypes.OperatorIdFromKeyPair(blsKeyPair)
+	if operatorId != calculatedOperatorId {
+		return nil, fmt.Errorf("the operator Id calculated from BLS public key does not match the on chain operator Id")
+	}
```
MegaRedHand:
```suggestion
	computedOperatorId := sdktypes.OperatorIdFromKeyPair(blsKeyPair)
	if operatorId != computedOperatorId {
		return nil, fmt.Errorf("the operator ID computed from the BLS keypair does not match the on-chain one")
	}
```

PR #737, opened 2025-06-16
feat: Add grpc logging interceptors

ian-shim:
### What Changed?

Add grpc interceptors for each logger

### Reviewer Checklist
- [ ] Code is well-documented

### DIFF
--- a/logging/interceptors.go
+++ b/logging/interceptors.go
@@ -1,0 +1,41 @@
+package logging
+
+import (
+	"context"
+
+	grpclogging "github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/logging"
+)
+
+func (s SLogger) InterceptorLogger() grpclogging.Logger {
+	return grpclogging.LoggerFunc(func(ctx context.Context, lvl grpclogging.Level, msg string, fields ...any) {
+		switch lvl {
+		case grpclogging.LevelDebug:
+			s.Debug(msg, fields...)
+		case grpclogging.LevelInfo:
+			s.Info(msg, fields...)
+		case grpclogging.LevelWarn:
+			s.Warn(msg, fields...)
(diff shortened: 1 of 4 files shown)


logging/logger.go
```diff
@@ -22,4 +24,6 @@ type Logger interface {
 	Fatalf(template string, args ...interface{})
 
 	With(tags ...any) Logger
+
+	InterceptorLogger() grpclogging.Logger
```
samlaf, line 28: Don't think this belongs here. Following https://neugierig.org/software/blog/2019/11/interface-pattern.html), the main (only) reason for defining this interface is such that it can be used inside sdk functions, such that clients of the eigensdk can pass in their own logger (which could be different from slog or zap). 

Since we aren't using InterceptorLogger in any eigensdk function, then the function doesn't belong in this interface. Otherwise we are potentially breaking clients (that were passing their logger which doesn't implement InterceptorLogger), and now they are forced to implement this random grpc method which has nothing to do with anything inside eigensdk.

If we need an interface with the InterceptorLogger method for eigenda, then that interface should be defined inside eigenda codebase imo.

logging/interceptors.go
```diff
@@ -0,0 +1,41 @@
+package logging
+
+import (
+	"context"
+
+	grpclogging "github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/logging"
+)
+
+func (s SLogger) InterceptorLogger() grpclogging.Logger {
```
samlaf, line 9: add documentation: what/how is this used for?

logging/logger.go
ian-shim: good point.
I moved it to eigenda repo: https://github.com/Layr-Labs/eigenda/pull/1690

